PR #170, opened 2021-03-26
CMake improvements: see if SDL2 target is defined; allow to build as a static library

eliasdaler:
Hello.
This patch fixes two problems that I've encountered with SDL_image.

1) It didn't allow me to build it as a static library - `add_library` was called with `SHARED` flag, so `BUILD_SHARED_LIBS` had no effect.
2) I build SDL_image with SDL2 in a "superbuild" way by doing this:

```cmake
add_subdirectory(sdl2_src)
add_subdirectory(sdl_image_src)
```

Unfortunately, `find_package` didn't allow me to do it, because SDL2 was not found system-wide (it's not built when I call add_subdirectory). However, if you don't call `find_package` if `SDL2` target is present, this will work, because SDL2 will be compiled later and will successfully link with SDL_image afterwards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,7 @@
 cmake_minimum_required(VERSION 3.3.2)
 project(SDL_image C)
 
-if (ANDROID)
-
-else()
+if (NOT ANDROID AND NOT TARGET SDL2)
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.

PR #172, opened 2021-03-29
More CMake fixes

eliasdaler:
Hello. This patch is a lot more complex than my previous CMake one, but it's not possible to fix only one thing at a time, unfortunately.

Currently, the CMake build is broken in several places and here's what this patch does:

* Link to a correct SDL target depending on build type (static or shared)
* Link to a correct libpng target based on the build type
* Fix libjpeg build (`jmemansi.c` wasn't included in the build)
* Fix showimage build on Windows - need to link to SDL2main
* Use bundled zlib for build
* Fix webp build (it doesn't add needed include directory for some reason)
* Add zconf.h to gitignore - maybe it should be deleted from the repo even? It's generated during the build of zlib in the binary dir. And zlib's CMake renames `zconf.h` to `zconf.h.included` in the source directory, which is annoying and can't be prevented...

Dynamic linking doesn't work on Windows with `SUPPORT_WEBP=ON`, because libwebp doesn't define `WEBP_EXTERN` as `__declspec(dllexport)` - I didn't find a way to fix it yet, unfortunately (maybe the issue should be created so that we don't forget about it?)

P.S. In the next patches it will be good to add ability for user to link their own libpng, libz, libjpeg and libwebp if they want to. I currently build them myself and it's currently impossible to force SDL_image to link to them instead of bundled ones.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ Release
 *.suo
 *.sdf
 *.opensdf
+
+# generated zlib files
+external/zlib-1.2.11/zconf.h
+external/zlib-1.2.11/zconf.h.included
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 3.3.2)
 project(SDL_image C)
 
-if (NOT ANDROID AND NOT TARGET SDL2)
-    find_package(SDL2 REQUIRED)
+if (NOT ANDROID AND NOT (TARGET SDL2 OR TARGET SDL2-static))
+	find_package(SDL2 REQUIRED)
 endif()
 
 option(SUPPORT_JPG "Support loading JPEG images" ON)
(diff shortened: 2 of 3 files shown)


eliasdaler:
@rafaeldelboni thanks! 

I think that the better handling of zlib and libpng should be done in the next patch.
I want to add ability for user to link with system deps + the dependencies that can come from users manually doing `add_subdirectory` for zlib and libpng. The `FORCE` method inside SDL_Image won't be very nice for everyone, so I'll think about it too. But let's start with something simple - I think that this patch is the least invasive one.

@sezero, @icculus - can you please take a look at this PR?

icculus:
I’m of the opinion that I don’t think a great deal about the CMake projects in any of the external libraries like SDL_image, so if @sezero has no concerns, let’s merge this and any other CMake stuff Elias has done.

Any CMake changes to SDL itself, though, I want to personally inspect before merging (at least for now).

sezero:
> if @sezero has no concerns, let’s merge this and any other CMake stuff Elias has done.

I don't like cmake, nor do I have much experience with it, but after giving
a _very_ brief look at it, it should be OK.  Let's merge it, and if anything is
broken we can fix later.

PR #191, opened 2021-09-04
Make universal fat binaries on MacOS

ankith26:
Code just copied from main SDL repo, credits go to Ryan!


slouken:
Rather than copy this file to every project, we'll let people grab it from the main SDL repo.

PR #196, opened 2021-10-11
build: Respect environment LDFLAGS

tbaederr:
Prepend them to libSDL2_image_la_LDFLAGS. This fixes the build when
using clang and LTO.

Clang needs `-flto` in both the `CFLAGS` and `LDFLAGS`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -52,6 +52,7 @@ EXTRA_DIST =			\
 
 if USE_VERSION_RC
 libSDL2_image_la_LDFLAGS = 	\
+	$(LDFLAGS) \
 	-no-undefined		\
 	-release $(LT_RELEASE)	\
 	-version-info $(LT_CURRENT):$(LT_REVISION):$(LT_AGE) -Wl,version.o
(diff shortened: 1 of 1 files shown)


smcv:
This shouldn't be necessary in an Autotools build system, because the generated `Makefile.in` includes both the library's `libfoo_la_LDFLAGS` and the user-supplied `LDFLAGS` in the link line:

```
libSDL2_image_la_LINK = $(LIBTOOL) $(AM_V_lt) $(AM_LIBTOOLFLAGS) \
	$(LIBTOOLFLAGS) --mode=link $(OBJCLD) $(AM_OBJCFLAGS) \
	$(OBJCFLAGS) $(libSDL2_image_la_LDFLAGS) $(LDFLAGS) -o $@
```

so perhaps something else is going wrong?

Autotools build systems expect the `LDFLAGS` to either be passed to `configure` as an argument like `LDFLAGS=-flto`, placed in the environment at `configure` time like `LDFLAGS=-flto ./configure ...`, or passed to `make` as an argument like `make LDFLAGS=-flto`. Which one are you doing?

tbaederr:
This is an error that occurs when using the `%configure` macro in Fedora's rpm config. But now I think this is just a case of an old libtool stripping the `-flto`.

PR #197, opened 2021-10-11
Removed Generated zconf.h

HumanGamer:
This file is already ignored but was left in by mistake. It gets quite annoying as I have SDL_image as a submodule and SourceTree keeps complaining that the project is changed as building with CMake deletes this file.


sezero:
@slouken: This will make android builds to fail: I suggest reverting.

PR #199, opened 2021-12-05
Add QOI format support 

hovertank3d:
https://github.com/phoboslab/qoi


sezero:
If @slouken or @icculus wants this, this patch needs changes:

- https://github.com/phoboslab/qoi/blob/master/README.md has scary
  warnings a.out the project's readiness
- qoi.h is not buildable in c90 mode: See the attached patch which
  makes it build in c90 mode.
- None of the libc/libm func redefines in IMG_qoi.c are needed:
  We only need to define QOI_MALLOC and QOI_FREE.
- We don't need qoi.h stdio functionality in IMG_qoi.c: we should
  define QOI_NO_STDIO.
- qoi.h stores ftell result as `int` but it should be `long`,
  rest of its api (e.g. qoi_encode) relies on an int size ptr.
  That's minor, but wanted to mention anyway.
- autotools changes are wrong: You should only touch Makefile.am
  and configure.ac, the rest I can handle regenerating here.
- VisualC and Xcode projects need updating for the new addition.

Here are the patches I used:

- patch for IMG_qoi.c: [patch-IMG_qoi.txt](https://github.com/libsdl-org/SDL_image/files/7656405/patch-IMG_qoi.txt)
- patch for qoi.h: [patch-qoi_h.txt](https://github.com/libsdl-org/SDL_image/files/7656406/patch-qoi_h.txt)
- patch for autotools (excludes generated files): [patch-autotools.txt](https://github.com/libsdl-org/SDL_image/files/7656408/patch-autotools.txt)

sezero:
Additional note: My gcc emits the following warnings:
```
qoi.h: In function ‘qoi_encode’:
qoi.h:356: warning: missing braces around initializer
qoi.h:356: warning: (near initialization for ‘index[0]’)
qoi.h: In function ‘qoi_decode’:
qoi.h:497: warning: missing braces around initializer
qoi.h:497: warning: (near initialization for ‘index[0]’)
```
These are `qoi_rgba_t index[64] = {0};` and a compiler (e.g. MSVC)
can optimize those to memset(index,0,sizeof(x)) which can be a
problem in SDL's MSVC builds (SDL uses its own `SDL_memset` proc.)
I suggest that qoi project should expose a `QOI_ZERO` macro and
use it instead of those initializers, if it does that then SDL can
use its own `SDL_zeroa` macro for it.

sezero:
FWIW, I submitted my proposed qoi.h changes to mainstream at https://github.com/phoboslab/qoi/pull/67

slouken:
@sezero, thanks for the review. The README states 12/20/2021 as the date that the specification will be finalized, so let's wait for that before applying this patch. Hopefully the upstream patches will have landed as well.

sezero:
And here is a patch to address zero-initializer issue I mentioned above (also submitted to upstream):
[patch-zero-initializer.txt](https://github.com/libsdl-org/SDL_image/files/7662793/patch-zero-initializer.txt)

sezero:
The patches I suggested to qoi mainstream are in, so I updated this PR
to reflect them, along with the build system updates I requested. MSVC
and Xcode project files still need updating, by someone other than me.

@slouken: The rest is up to you.

sezero:
The format seems to have finalized: https://github.com/phoboslab/qoi/issues/95

I updated the pull request a final time.

PR #201, opened 2021-12-14
gif: Don't import internal function IMG_LoadGIF_RW_Internal

smcv:
This isn't mentioned in the header file and it returns a type that
callers cannot understand without knowing internal implementation
details, so I assume it was meant to be private to SDL_image.

### DIFF
--- a/IMG_gif.c
+++ b/IMG_gif.c
@@ -199,7 +199,7 @@ static SDL_bool NormalizeFrames(Frame_t *frames, int count)
     return SDL_TRUE;
 }
 
-Anim_t *
+static Anim_t *
 IMG_LoadGIF_RW_Internal(SDL_RWops *src, SDL_bool load_anim)
 {
     unsigned char buf[16];


slouken:
Symbol visibility should default to private as well. Is that not the case here?

sezero:
It _is_ internal: it is not in public header and not marked with `DECLSPEC`
Thanks.

smcv:
> Symbol visibility should default to private as well. Is that not the case here?

The core SDL library has explicit visibility control, and so does SDL_mixer; but SDL_image, SDL_net and SDL_ttf don't seem to have that feature yet.

smcv:
> It _is_ internal: it is not in public header and not marked with `DECLSPEC` Thanks.

That might be enough to make it private on Windows, but it's exported on at least Linux when building with Autotools.

sezero:
Oh hell.. Their configury really doesn't add `-fvisibility=hidden`. Will make them so.

smcv:
> Their configury really doesn't add -fvisibility=hidden. Will make them so.

Thanks - that'll be a big improvement from a downstream distribution point of view!

sezero:
Quick lazy question: Is returning from a function (like below) legal
in all sh implementations other than bash?
```diff
diff --git a/configure.ac b/configure.ac
index 3cf2dc9..62ecf8f 100644
--- a/configure.ac
+++ b/configure.ac
@@ -157,6 +157,14 @@ dnl  Details of this flag are here: http://gcc.gnu.org/wiki/Visibility
 CheckVisibilityHidden()
 {
     AC_MSG_CHECKING(for GCC -fvisibility=hidden option)
+
+    case "$host" in
+    *-*-cygwin* | *-*-mingw* | *-*-os2*)
+        AC_MSG_RESULT([ignored for $host_os])
+        return
+        ;;
+    esac
+
     have_gcc_fvisibility=no
 
     visibility_CFLAGS="-fvisibility=hidden"
```

smcv:
> Is returning from a function (like below) legal in all sh implementations other than bash?

I believe so. dash has it, and POSIX [documents it](https://pubs.opengroup.org/onlinepubs/009695399/utilities/return.html).

sezero:
Thanks

sezero:
Touched all of the satellite projects' autotools for visibility attributes.

If someone wants the same for cmake (not mature or present in many,
if not all of those projects), then someone else should do it.

PR #203, opened 2021-12-15
Add option 'SDL_IMAGE_USE_WIC_BACKEND' to cmakelists

XiangYyang:
I noticed some details in the cmakelists.txt and a missing option "SDL_IMAGE_USE_WIC_BACKEND". I added them.

 1. Add prefix: "SDL_IMAGE_XXXX..."
 2. Add option "SDL_IMAGE_USE_WIC" to enable WIC on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,11 +9,20 @@ option(SDL_IMAGE_SUPPORT_JPG "Support loading JPEG images" ON)
 option(SDL_IMAGE_SUPPORT_PNG "Support loading PNG images" ON)
 option(SDL_IMAGE_SUPPORT_WEBP "Support loading WEBP images" OFF)
 option(SDL_IMAGE_BUILD_SHOWIMAGE "Build the showimage sample program" OFF)
+option(SDL_IMAGE_USE_WIC "Use WIC Backend on Windows" OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -5,15 +5,24 @@ if (NOT ANDROID AND NOT (TARGET SDL2 OR TARGET SDL2-static))
...
 if (APPLE)
 	option(ENABLE_APPLE_IMAGEIO "Use native Mac OS X frameworks for loading images" ON)
 endif()
 
+if (NOT WIN32 AND SDL_IMAGE_USE_WIC)
+	message(FATAL_ERROR "WIC is only supported on Windows")
+endif()
+if (NOT SDL_IMAGE_SUPPORT_PNG OR NOT SDL_IMAGE_SUPPORT_JPG)
+	if (SDL_IMAGE_USE_WIC)
+		message(FATAL_ERROR "WIC requires PNG and JPG support.")
```
sezero, line 23: This is wrong. If you really want, you can add to the description that this option provides builtin support for jpg, png and tiff.

XiangYyang: Oh, that's right. I forgot it. However, I think that the maintain for using WIC is used to reduce third-party dependency, in this library, we don't seem to need to add a new option for tiff format — 
it will not be compiled, when using WIC.

PR #222, opened 2022-05-05
Use minor version for feature releases

smcv:
Requested by @slouken, and similar to libsdl-org/SDL#5602.

Depends on (and includes) #223, because I want to have some sort of CI before making structural changes like this (and in particular, it's important that `./test-versioning.sh` gets run regularly).

* build: Use minor version number for feature releases
    
    Similar to what has been done in SDL itself, this allows micro versions
    with an odd minor version to be used as prereleases for a stable
    release. It also allows micro versions with an even minor version to
    be used to fix regressions in their corresponding .0 release
    if that becomes necessary.

* cmake: On Unix, make the library ABI-compatible with the Autotools build
    
    This is not yet implemented for macOS, only for generic Unix (in practice
    meaning Linux and *BSD).


slouken:
> I used 2.5.0 as the new development version number in an attempt to indicate continuity from the old 2.0.x versions (so the stable versions suitable for Linux distro use will go 2.0.4 -> 2.0.5 -> 2.6.0 -> 2.8.0), but it would also be OK to use 2.7.0 or 2.1.0 or something - let me know what you'd prefer.

Yep, that sounds good.

smcv:
After merging this, please could someone with the right environment re-run `./autogen.sh` and commit `configure`? Sorry, I don't know what operating system the SDL maintainers use for the "official" `configure` script.

I could do the `autoreconf` myself, but that would come from Debian testing/unstable, which is newer, leading to a huge and unreviewable diff.

The usual convention is for generated files like `configure` to *not* be committed to the repository, so that users have to either download a `make dist` tarball, or run `autogen.sh` themselves if they are working from git - but I realise that would make it harder to do the simplest possible checkout-and-build.

SDL2 will need a similar treatment.

sezero:
I just regenerated autofoo and pushed

PR #223, opened 2022-05-05
Add Github CI workflows

smcv:
Depends on #222 (but could be adapted to not).


smcv:
Do Github workflows need to be enabled on a per-repository basis?

smcv:
> Do Github workflows need to be enabled on a per-repository basis?

Looks like yes they do, but I have them enabled in my fork, so I can use https://github.com/smcv/SDL_image/tree/workflows to test this.

smcv:
I think this is ready now. @slouken, would it be possible to enable Github Actions on this repo?

For now I've been testing this in my fork: https://github.com/smcv/SDL_image/actions

slouken:
We should be good to go, actions are enabled on this repo.

PR #225, opened 2022-05-05
build: Pass in the build system's version info as cpp defines

smcv:
This allows the static assertions in IMG.c to assert that the build
system agrees with SDL_image.h.


sezero:
I do **hate** autoconf adding `-DPACKAGE???=xxx` crapola to CFLAGS.
We don't use them, they may be harmless but they add a lot of garbage
to our cmdline and make the generated makefile harder to browse when
needed. Is there any way to suppress their generation?

(Irrelevant to your PR, but there I said it...)

smcv:
If we let Autoconf generate a `config.h`, then the `PACKAGE`, `HAVE_FOO`, etc. will be in there, rather than on the command-line (and we could either add `-include config.h` to the command-line, or systematically add `#include "config.h"` to each `.c` file). Would that help?

sezero:
Nope..

Let's get this in as it is.

PR #232, opened 2022-05-12
Don't link tinyjpeg if we have a hard dependency on libjpeg

smcv:
Linux distributions like to control the number of image loading/saving
implementations they are compiling, because each one comes with a risk
of security vulnerabilities.

If we've built with --disable-stb-image and --disable-jpg-shared
(ordinary linking with -ljpeg, as opposed to using dlopen()) then
initializing libjpeg can't fail: either it's present and works, or
loading SDL_image would have failed in the runtime linker. As a result,
there's no point in having the code path that can fall back to tiny_jpeg
in this configuration.

### DIFF
--- a/IMG_jpg.c
+++ b/IMG_jpg.c
@@ -679,7 +679,8 @@ SDL_Surface *IMG_LoadJPG_RW(SDL_RWops *src)
 
 #endif /* LOAD_JPG */
 
-#if SAVE_JPG
+/* Use tinyjpeg as a fallback if we don't have a hard dependency on libjpeg */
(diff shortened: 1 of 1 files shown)


smcv:
Tested by inserting `#error Should not be used` into `stb_image.h` and `tiny_jpeg.h`. As expected, the build continues to succeed with `--disable-stb-image --disable-jpg-shared`, but fails otherwise.

sezero:
This is in.

sezero:
I guess IMG_png.c needs something similar for miniz??

sezero:
> That looks right to me (with your edits)

Pushed as https://github.com/libsdl-org/SDL_image/commit/686ad262141126d5cd40fd71f368eab1bbec14aa

PR #237, opened 2022-05-16
cmake: fixed the use of SDL_image with disabled BUILD_SHARED_LIBS

rmg-nik:
- Fixed using SDL_image as subproject with cmake's add_subdirectory with disabled BUILD_SHARED_LIBS
- Removed checks for ANDROID when detecting SDL2 or SDL2-static

From now we can use SDL_image with cmake in this way

```
option(BUILD_SHARED_LIBS "Build the library as a shared library" OFF)
option(BUILD_SAMPLES "Build the sample program(s)" OFF)
option(SDL_STATIC_PIC "Static version of the library should be built with Position Independent Code" ON)

add_subdirectory(external/SDL)
add_subdirectory(external/SDL_image )
add_subdirectory(path_to_your_project)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ project(SDL2_image LANGUAGES C
 
 message(STATUS "Configuring ${PROJECT_NAME} ${PROJECT_VERSION}")
 
-if (NOT ANDROID AND NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
+if (NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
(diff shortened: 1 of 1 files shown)


madebr:
FYI, this fix is not 100% correct since when building a static SDL2 + SDL2_image, SDL2_image should link to SDL2::SDL2-static instead of `SDL2::SDL2`.
This will soon be rectified.

madebr:
@rmg-nik, see #238 for my take on it.
It also adds a `SDL2_image::SDL2_image-static` target when building static.

rmg-nik:
@madebr Got it, thanks for the fix.

Issue #235, opened 2022-05-14
Define for SAVE_PNG support


1bsyl:
It would be nice to re-add the define to compile out the PNG support: 

eg:

```diff
diff --git a/Android.mk b/Android.mk
index c6d83db..7e9ab1c 100644
--- a/Android.mk
+++ b/Android.mk
@@ -127,6 +136,11 @@ ifeq ($(SUPPORT_PNG),true)
     LOCAL_CFLAGS += -DLOAD_PNG
     LOCAL_STATIC_LIBRARIES += png
     LOCAL_LDLIBS += -lz
+ifeq ($(SUPPORT_SAVE_PNG),true)
+    LOCAL_CFLAGS += -DSDL_IMAGE_SAVE_PNG=1
+else
+    LOCAL_CFLAGS += -DSDL_IMAGE_SAVE_PNG=0
+endif
 endif
 
 ifeq ($(SUPPORT_WEBP),true)
diff --git a/IMG_png.c b/IMG_png.c
index 462fe2f..61b5621 100644
--- a/IMG_png.c
+++ b/IMG_png.c
@@ -23,9 +23,8 @@
 
 #include "SDL_image.h"
 
-/* We'll have PNG save support by default */
-#ifndef SAVE_PNG
-#define SAVE_PNG    1
+#if !defined(SDL_IMAGE_SAVE_PNG)
+#  define SDL_IMAGE_SAVE_PNG 1
 #endif
 
 #if defined(USE_STBIMAGE)
@@ -115,7 +116,7 @@ static struct {
     jmp_buf* (*png_set_longjmp_fn) (png_structrp, png_longjmp_ptr, size_t);
 #endif
 #endif
-#if SAVE_PNG
+#if SDL_IMAGE_SAVE_PNG
     png_structp (*png_create_write_struct) (png_const_charp user_png_ver, png_voidp error_ptr, png_error_ptr error_fn, png_error_ptr warn_fn);
     void (*png_destroy_write_struct) (png_structpp png_ptr_ptr, png_infopp info_ptr_ptr);
     void (*png_set_write_fn) (png_structrp png_ptr, png_voidp io_ptr, png_rw_ptr write_data_fn, png_flush_ptr output_flush_fn);
@@ -170,7 +173,7 @@ int IMG_InitPNG()
         FUNCTION_LOADER(png_set_longjmp_fn, jmp_buf* (*) (png_structrp, png_longjmp_ptr, size_t))
 #endif
 #endif
[...]

PR #238, opened 2022-05-16
Improve CMake build system

madebr:
- Use 'SDL2IMAGE_' prefix for all options
- Create SDL2_image::SDL2_image when building a shared library,
  create SDL2_image::SDL2_image-static when building a static library
- Aim is to use same CMake layout as SDL_ttf/SDL_mixer
- Create libSDL2_imaged.so symbolic link ('d' suffix) (when building debug build type)
- Add PrivateSdlFunctions.cmake script for common functionality between SDL_ttf/SDL_mixer
- Add FindXXX.cmake scripts for dependencies that don't provide cmake configuration files 
- Bring jpeg/png saving support to cmake (#235)
- Update the git submodules
- Test CMake on mingw64 (using system dependencies there)
- Add MacOS CI
- Create symbolic link at build time + install it
- Only install SDL2_image.pc when building a shared SDL2_image (to avoid overwriting when later installing a static SDL2_image)
- Allow parallel installation of cmake support of a static and shared SDL2_image
- Add a single `SDL2IMAGE_VENDORED` option to use vendored libraries for everything, or none at all.
- Add Macos dylib versioning

I haven't added these scripts to `EXTRA_DIST` in `Makefile.am` yet, to allow some maturing + extra tests.

Need review on:
- ease of use: the cmake recipe is strict with its configuration options. When an option is configured to be enabled (`-DSDL2IMAGE_xxx=ON`), the cmake script will complain loudly instead of silently ignoring the option.
- vendored libraries: build these as shared or static libraries? Always building these as static libraries makes perhaps most sense.


.github/workflows/main.yml
```diff
@@ -32,13 +35,29 @@ jobs:
           ${{ matrix.platform.msys-env }}-cmake
           ${{ matrix.platform.msys-env }}-gcc
           ${{ matrix.platform.msys-env }}-libjpeg-turbo
+          ${{ matrix.platform.msys-env }}-libjxl
```
sezero, line 38: You can add `libavif` here, too.

slouken: We want to use the versions we ship with SDL_image, right?

sezero: Autofoo doesn't do that

madebr: You can switch between vendored and system packages using the `SDL2IMAGE_VENDORED` option.
There exist package managers on all platforms that would make us vendoring the 3rd party libraries not necessary anymore.
It would also make the cmake script much cleaner.

sezero: > You can switch between vendored and system packages using the `SDL2IMAGE_VENDORED` option.

Cmake: yes. Autotools doesn't do that , as far as I know.

madebr: > You can add `libavif` here, too.

We need libavif 0.10
https://github.com/madebr/SDL_image/runs/6479514233?check_suite_focus=true#step:7:25

sezero: > We need libavif 0.10

Do we really, though? 0.10 is what we forked from mainstream, can't we
run against something somewhat older? (@slouken?)

madebr: The error is very benign, 0.10 added `strictFlags`:
```
/home/maarten/programming/SDL_image/IMG_avif.c:257:12: error: ‘avifDecoder’ has no member named ‘strictFlags’
  257 |     decoder->strictFlags = AVIF_STRICT_DISABLED;
      |            ^~
/home/maarten/programming/SDL_image/IMG_avif.c:257:28: error: ‘AVIF_STRICT_DISABLED’ undeclared (first use in this function)
  257 |     decoder->strictFlags = AVIF_STRICT_DISABLED;
```

sezero: Then how did the autotools build succeed in that CI run?

sezero: According to git blame, strict flags was in v0.9.1:
https://github.com/AOMediaCodec/libavif/commit/9b942c24d3c3cb0547d79e253d013da7542edad0

madebr: Thanks, I will update the minimum version then.
I tested on a Linux Distribution with 0.9.0 and did not look up the minimum libavif version required.

madebr: I can't add libavif-dev for Ubuntu, as it's for Ubuntu 21.10+ only: https://packages.ubuntu.com/search?keywords=libavif-dev

sezero: > I can't add libavif-dev for Linux, as it's for Ubuntu 21.10+ only:

That's OK, we'll have something to test-build with mingw at least.

madebr: `${{ matrix.platform.msys-env }}-libavif` has been added.

smcv:
> vendored libraries: build these as shared or static libraries? Always building these as static libraries makes perhaps most sense.

I think so. The possible build configurations for a library (let's say libpng) are:

* use the system libpng
    * as a static library
    * as a shared library, linked with `-lpng` to be a mandatory dependency
    * as a shared library, dynamically loaded with `dlopen()` at runtime to be a "weak" dependency
* use the vendored libpng
    * as a static library
    * as a shared library, linked with `-lpng` to be a mandatory dependency
    * as a shared library, dynamically loaded with `dlopen()` at runtime to be a "weak" dependency

But if you're using the vendored libpng, then I don't think the two shared-linking options really make sense: you're paying the compile time and disk space cost of a copy of libpng *anyway*, whether it's shared (a separate file) or static (copied into the SDL_image library) . Having it as a shared library seems like it just brings more reasons why the overall system can fail, for instance conflicts with a system copy that happens to have the same SONAME.

In particular, building your own vendored shared library for *zlib* seems like a recipe for disaster, because on Linux and macOS there are likely to be OS components that depend on the OS's zlib. If you release a game that includes SDL_image, then wait 5 years, the OS zlib will probably be newer than your bundled zlib, so other OS components will expect to find that newer version, and will crash if they find a vendored shared library instead.

I like the way SDL_ttf handles this: if it's building a vendored Freetype or Harfbuzz, then the vendored copy is statically linked into the SDL_ttf library. This makes the SDL_ttf library larger than it would be when using an OS copy, but avoids conflicting with an older or newer system copy of Freetype or Harfbuzz.

SDL_ttf also uses `C_VISIBILITY_PRESET "hidden"` to make sure it doesn't accidentally export symbols from its vendored libraries, making them into merely an implementation detail.

smcv:
I'm not a maintainer, but:

I think if there are things that are functionally necessary (like being able to enable JPEG and PNG saving), it might be good to carve those out into smaller PRs, which will be quicker to review and less likely to conflict with other parallel work like #242.

Major refactoring and stylistic changes are basically guaranteed to conflict with everything else that touches the same files, so it's good to do those when the same files are otherwise quiet.

smcv:
> Major refactoring and stylistic changes are basically guaranteed to conflict with everything else that touches the same files, so it's good to do those when the same files are otherwise quiet.

When I've done this sort of refactor in other projects, the way I generally do it is to choose the piece of refactoring that I think is highest-priority or most likely to be accepted quickly, do that as its own PR, get it reviewed, and iterate until everything I want has gone in.

slouken:
@madebr, I'd like to make an SDL_image release candidate soon. I think this is the only work outstanding before I can do that. Do you have an ETA of having everything settled and ready for public eyes?

smcv:
Is this a blocker for a SDL_image release candidate? At the moment the source code releases aren't going to include `CMakeLists.txt` anyway.

slouken:
> vendored libraries: build these as shared or static libraries? Always building these as static libraries makes perhaps most sense.

For SDL_ttf, using FreeType and HarfBuzz as static libraries makes sense because they are core pieces of the library functionality. For SDL_image and SDL_mixer, the codecs are all optional pieces that can be used, or not, depending on an application's needs. 

For pre-built binary redistributables like DLLs for Windows or Frameworks for macOS, I like distributing the dependencies as optional shared libraries that can be present or not, depending on the application they're shipped with.

For building with Linux distributions, it makes sense to build in all supported codecs with dependencies on other codec libraries in the system. Distributions have often chosen to depend on those packages rather than have them be optional.

For building with CMake, this could either be a developer building a custom library for their own needs, or a Linux distribution choosing to use cmake instead of autotools. In the former case, it probably makes sense to use vendored libraries linked statically and defaulting off. In the latter case, the distribution maintainers will set whatever flags makes sense for the distribution.

For Android.mk, which is the build system most closely analogous to CMake, I've chosen to default external dependencies off, and include them as static libraries when they are enabled. I will probably do the same for SDL_mixer for the next release.

slouken:
> Is this a blocker for a SDL_image release candidate? At the moment the source code releases aren't going to include `CMakeLists.txt` anyway.

Yes, I'd like the next SDL library releases to have the improved CMake support.

smcv:
#246 sets up the vendored libraries to be built as static libraries *if* the `_SHARED` option is not enabled. Does that make sense?

madebr:
> #246 sets up the vendored libraries to be built as static libraries _if_ the `_SHARED` option is not enabled. Does that make sense?

This pr does the same.
It makes sense to not let the number of configurable options explode.

madebr:
I'm going to mark this pr as finished so other work can be done on top of it.

.github/workflows/main.yml
```diff
@@ -70,23 +91,23 @@ jobs:
...
-      run: cmake --build build/ --config Release
+      run: cmake --build build/ --config Release --verbose
     - name: Install
       if: "matrix.platform.shell == 'sh' && matrix.platform.cmake"
       run: |
         set -eu
         rm -fr DESTDIR-cmake
         DESTDIR=$(pwd)/DESTDIR-cmake cmake --install build/ --config Release
-        ( cd DESTDIR-cmake; find ) | LC_ALL=C sort -u
+        find DESTDIR-cmake | LC_ALL=C sort -u
```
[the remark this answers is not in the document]
madebr: Please revert this change in one of your (current) pr's. 
The reason I did this was because I wrongly though it failed on Macos here.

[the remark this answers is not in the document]
smcv: Will do. macOS might want `( cd ...; find . )` if its find(1) is less do-what-I-mean than Linux (I'm never sure what subset of the CLI is portable).

PR #241, opened 2022-05-17
showimage: Restore automation-friendly behaviour of -quit

smcv:
I added the -quit option for simple automated testing (with hindsight,
perhaps "-quit-after-loading" would have been a better name). The
intention was that something like

    showimage -quit -save out.bmp in.jpg -save out2.bmp in.png

would go through the motions of loading each image, save it to the
corresponding output file, but not actually wait for input or rendering.

---

I've started trying to learn the SDL_test framework well enough to add an automated test with better coverage, but that PR isn't ready yet, and probably shouldn't be considered a blocker for 2.5.1.

### DIFF
--- a/showimage.c
+++ b/showimage.c
@@ -56,6 +56,7 @@ int main(int argc, char *argv[])
     Uint32 flags;
     int i, w, h;
     int done = 0;
+    int quit = 0;
     SDL_Event event;
     const char *saveFile = NULL;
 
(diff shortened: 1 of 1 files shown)


madebr:
I added a `-timeout` argument to `showimage` to allow testing various formats on CI.
https://github.com/madebr/SDL_image/commit/a7a20749962eef613546d1513d5e860a9c4f7953

In the commit, I added it as a separate step in github CI, but it can be integrated into ctest/SDL_test.

PR #242, opened 2022-05-19
Add an automated test

smcv:
* Add a simple automated test
    
    This loads and saves various supported formats, and asserts that the
    same pixels (or close enough) end up in the output files.
    
    For a distribution that knows it intends to support particular formats,
    if an environment variable like SVG_IMAGE_TEST_REQUIRE_LOAD_JPG or
    SVG_IMAGE_TEST_REQUIRE_SAVE_JPG is set, the test will fail if the given
    format is not available.

* workflows: Run the new automated test for Autotools
    
    I haven't set it up for CMake yet (although it does work) because that
    would conflict pretty badly with #238.

---

The idea is that in Debian, I would set (for example) `SVG_IMAGE_TEST_REQUIRE_LOAD_JPG=1` and `SVG_IMAGE_TEST_REQUIRE_SAVE_JPG=1`, because I know that *for the build in Debian*, JPEG should always be supported, even if *for a generic upstream build* we can't guarantee that JPEG support will be enabled.


smcv:
https://github.com/smcv/SDL_image/tree/automated-test-cmake is the CMake build system glue for this, but will conflict with #238.

I think ideally we perhaps want the CMake test to build SDL_image as a shared library, but build the vendored dependencies as static libraries, since that's what someone who is vendoring the dependencies would probably want to do? But I'm not sure how to achieve that (or even whether it's possible without #238).

madebr:
Don't let my cmake pr block this one.
I'm currently working on SDL_mixer on and off.
Your suggestion of always building vendored dependencies as static libraries sounds good. It would also mean disabling dynamic loading.

Issue #243, opened 2022-05-19
CMakeLists.txt never enables saving PNG/JPEG


smcv:
CMakeLists.txt has options to enable *loading* lots of formats, but never enables the `#define` flags to *save* anything (currently JPEG and PNG are the only formats that can be saved).

smcv:
I think #238 might already have a fix for this.

smcv:
Fixed by #238.

PR #244, opened 2022-05-20
cmake: Add build options for whether to support saving PNGs

smcv:
Based on code from @madebr.

Resolves: https://github.com/libsdl-org/SDL_image/issues/243


madebr (approved):

smcv:
I know @madebr wants to rename all the CMake build options, but in the interests of minimal changes, this PR keeps them consistent with the short names used in `Android.mk`.

madebr:
FYI, the reason I'm renaming the options is to avoid accidental mixing with completely independent projects.
Also, CMake configuration tools such as cmake-gui/ccmake group options by their prefix.

smcv:
> FYI, the reason I'm renaming the options is to avoid accidental mixing with completely independent projects.

Sure, that makes sense, I just don't want to get entangled with larger changes like that one right now.

smcv:
After this and #246 both land, the test glue in #246 can have this change:

```diff
-        # TODO: https://github.com/libsdl-org/SDL_image/issues/243
-        export SDL_IMAGE_TEST_REQUIRE_SAVE_JPG=0
-        export SDL_IMAGE_TEST_REQUIRE_SAVE_PNG=0
+        export SDL_IMAGE_TEST_REQUIRE_SAVE_JPG=1
+        export SDL_IMAGE_TEST_REQUIRE_SAVE_PNG=1
```

smcv:
Obsoleted by #238

PR #245, opened 2022-05-20
cmake: Test static build in CI

smcv:
* cmake: Use a shared SDL library if that's all we have

* workflows: Exercise static linking with CMake
    
    There are enough subtleties with the vendored libraries that it seems
    worthwhile to test this.

* workflows: Build (but don't run) the automated test for CMake builds
    
    We can't run this yet, because the vendored libraries won't be in the
    dlopen() search path.


test/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ foreach(prog ${ALL_TESTS})
         target_link_libraries(${prog} PRIVATE SDL2::SDL2main)
     endif()
     target_link_libraries(${prog} PRIVATE SDL2_test)
-    if (BUILD_SHARED_LIBS)
+    if (BUILD_SHARED_LIBS OR NOT TARGET SDL2::SDL2-static)
```
madebr, line 40: (not related to this change)
Perhaps, it makes sense to set a timeout to avoid an accidental stuck test:
https://cmake.org/cmake/help/latest/prop_test/TIMEOUT.html

smcv: Perhaps, but his branch doesn't run any tests in the CMake build anyway. (#246 does.)

smcv: #247

madebr: Thanks!

FYI, the tests are failing on Macos: https://github.com/madebr/SDL_image/runs/6528435637?check_suite_focus=true

smcv: #248

smcv:
Superseded by later PRs.

PR #246, opened 2022-05-20
cmake: Build vendored libraries statically if SHARED not enabled

smcv:
Depends on #245 

* cmake: Build vendored libraries statically if SHARED not enabled
    
    If we're building vendored libraries, we'll often want them to be an
    implementation detail of SDL_image that will not conflict with system
    libraries (or even with libraries vendored by a different dependency),
    which might be a different version.

* cmake: Add a default for whether to use shared linking
    
    As with vendored libraries, if we want to use one setting or the other
    for one of the dependencies, it's likely that we'll want to do the same
    for all of the dependencies.

* workflows: Enable the automated test for CMake builds
    
    If we link the vendored libraries statically, then we should have
    everything we need to make this test work.


smcv:
Superseded by #238 and #249.

PR #249, opened 2022-05-20
More automated testing

smcv:
* workflows: Collect artifacts for all jobs, not just Linux

* workflows: Exercise static linking with CMake
    
    There are enough subtleties with the vendored libraries that it seems
    worthwhile to test this.

* cmake: Add a shortcut for setting most dependencies _SHARED or not

* workflows: Run the automated test for CMake builds
    
    We can't use dlopen() for the dependencies, because they won't be in
    the search path at runtime, so link them statically here.


smcv:
Ugh, sorry, I should have marked that as a draft - it's somewhat broken.

slouken:
Oops. I'll ignore test failures for the moment then.

.github/workflows/main.yml
```diff
@@ -93,21 +94,62 @@ jobs:
...
+        export SDL_IMAGE_TEST_REQUIRE_SAVE_JPG=1
+        export SDL_IMAGE_TEST_REQUIRE_SAVE_PNG=1
+
+        ctest -VV --test-dir build/test
     - name: Install
       if: "matrix.platform.shell == 'sh' && matrix.platform.cmake"
       run: |
         set -eu
         rm -fr DESTDIR-cmake
         DESTDIR=$(pwd)/DESTDIR-cmake cmake --install build/ --config Release
```
madebr, line 141: Additionally, I would create a tarball of the install prefix, and add it to the artifacts (for both autotools and cmake).

CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ option(CMAKE_POSITION_INDEPENDENT_CODE "Build static libraries with -fPIC" ON)
 option(BUILD_SHARED_LIBS "Build the library as a shared library" ON)
 
 option(SDL2IMAGE_INSTALL "Enable SDL2_image install target" ON)
+option(SDL2IMAGE_DEPS_SHARED "Load dependencies dynamically" ON)
```
madebr, line 50: This is one of the things I was not sure about in the cmake build script:
does it make sense to dynamically load only certain dependencies?

Currently, this variable is used as a default variable for `cmake_dependent°option`.
Personally, I would ditch all `SDL2IMAGE_xxx_SHARED` options and only use this one variable.

smcv: > does it make sense to dynamically load only certain dependencies?

I don't see why not: a Linux distribution that is less hardline about normal linking than Debian might want to make libjpeg and libpng be hard dependencies, but make libavif and libjxl weak/optional dependencies.

.github/workflows/main.yml
smcv: For what purpose? These binaries are not really expected to work on any system other than the one they were built on (which is essentially arbitrary).

smcv:
> Ugh, sorry, I should have marked that as a draft - it's somewhat broken.

#251 resolves that, and I'll have another try another time. Sorry about that.

Now that we're starting to have meaningful CI, it's probably best to wait for it to pass before hitting merge.

.github/workflows/main.yml
madebr: Convenience binaries for allowing people testing latest master without them needing to rebuild SDL_image.
But your reasoning sounds solid.

PR #250, opened 2022-05-20
workflows: Don't try to build tests on mingw + cmake

smcv:
The mingw-w64-*-SDL2 packages don't contain the SDL_test library,
so skip the tests on Windows, like we already did for Autotools.

---

This should hopefully fix the mingw64+cmake build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,22 +91,26 @@ jobs:
     - name: Configure CMake
       if: "matrix.platform.cmake"
       run: |
-        cmake -B build \
+        set -- \
           -DBUILD_SHARED_LIBS=ON \
           -DCMAKE_BUILD_TYPE=Release \
           -DSDL2IMAGE_DEPS_SHARED=OFF \
           -DSDL2IMAGE_SAMPLES=ON \
-          -DSDL2IMAGE_TESTS=ON \
-          -DSDL2IMAGE_TESTS_INSTALL=ON \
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -91,22 +91,26 @@ jobs:
...
       if: "matrix.platform.cmake"
       run: |
-        cmake -B build \
+        set -- \
           -DBUILD_SHARED_LIBS=ON \
           -DCMAKE_BUILD_TYPE=Release \
           -DSDL2IMAGE_DEPS_SHARED=OFF \
           -DSDL2IMAGE_SAMPLES=ON \
-          -DSDL2IMAGE_TESTS=ON \
-          -DSDL2IMAGE_TESTS_INSTALL=ON \
```
madebr, line 100: You can leave `-DSDL2IMAGE_TESTS_INSTALL=ON` here.
It will automatically be disabled when either `SDL2IMAGE_TESTS` or `SDL2IMAGE_INSTALL` are disabled (because of `cmake_dependent_option`).

smcv: Good thinking.

I'll come back to this when I have more time to investigate what's broken in the mostly-static CMake build.

smcv:
Superseded by #251 and then #258.

.github/workflows/main.yml
smcv: I followed this suggestion in #258.

Issue #248, opened 2022-05-20
Automated test failing on macOS


smcv:
When @madebr's branch added coverage on macOS, the automated test that I recently added is failing.

smcv:
@madebr, if you don't have a Mac where you can run this locally, you could set up the "Upload artifacts" workflow step to run on macOS as well as Linux. You'll get a zip file with the comparison images. Ideally the zip file should have a different name for each test run to avoid race conditions, which would probably mean adding a unique `artifacts_name` key to `matrix.platform` or something.

smcv:
AVIF loading is not quite the same as the reference:

```
2022-05-20 16:35:24.987 testimage[7044:21439] ERROR: 05/20/22 16:35:24: First detected occurrence at position 0,0 with a squared RGB-difference of 77.
```

I thought the decoder was deterministic; but if it isn't, then the thing to do is to look at the results, check that they're visually indistinguishable, and if they are, increase the `Format.tolerance` to some reasonable value like I did for JPEG.

smcv:
Loading `.cur` also doesn't match the reference:

```
2022-05-20 16:35:25.011 testimage[7044:21439] ERROR: 05/20/22 16:35:25: Comparison of pixels with allowable error of 0 failed 947 times.
2022-05-20 16:35:25.011 testimage[7044:21439] ERROR: 05/20/22 16:35:25: First detected occurrence at position 0,0 with a squared RGB-difference of 798.
```

That looks like a bigger difference - perhaps it's getting decoded incorrectly?

I suspect that decoding `.cur` is not a major use-case, so changing the `CUR` test-case to be `#if 0` or `#ifndef __darwin__` would be a reasonable workaround to get the rest of the tests running.

smcv:
Actually, a lot of these seem to be getting mis-decoded. Are we perhaps using a special Mac-specific loader implementation?

smcv:
```
2022-05-20 16:35:25.145 testimage[7044:21439] ERROR: 05/20/22 16:35:25: Assert 'TIF loading should be enabled': Failed
```

This one is easy: if the Mac build is not meant to have TIFF support, then we want to only `export SDL_IMAGE_TEST_REQUIRE_LOAD_TIF=1` on Linux.

slouken:
> Actually, a lot of these seem to be getting mis-decoded. Are we perhaps using a special Mac-specific loader implementation?

[...]

PR #251, opened 2022-05-20
Revert CI regressions

smcv:
* Fix name of Linux static build job

* Revert "workflows: Run the automated test for CMake builds"
    
    This reverts commit 72d9524536a34273210fab434d1c14f212ffd937.
    Building in this mode is failing; let's get the CI green again before
    fixing this.

* workflows: Disable automated test on macOS for now
    
    Various test failures are under investigation (see #248).


smcv:
@slouken: this makes the CI green again, and I'll have another try at enabling automated testing in CMake builds another time.

PR #252, opened 2022-05-20
test: Distribute template.test.in in source release tarballs

smcv:
We'll need this if we want to build with --enable-installed-tests
from an official source release, which I'd like to do in Debian.

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -20,7 +20,7 @@ AM_TESTS_ENVIRONMENT = \
 	$(NULL)
 
 EXTRA_DIST = \
-	CMakeLists.txt Makefile.os2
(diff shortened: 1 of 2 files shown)


sezero:
Please apply to test/Makefile.in too

smcv:
> Please apply to test/Makefile.in too

I'd been leaving that for people who have the right "blessed" build environment to update generated files, since I don't know what environment that is (it certainly doesn't match mine).

I've edited the generated file as requested, since this particular line maps relatively obviously from the source.

sezero:
>  since this particular line maps relatively obviously from the source.

Yes, that's exactly why I wanted it in the PR :)

PR #255, opened 2022-05-23
Use specific CMake version on github workflow

madebr:
The reasoning behind this extra github workflow dependency is enforcing compatibility with the minimum required cmake version we declared at the top of the cmake build script.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows (MSVC+CMake),         os: windows-latest, shell: sh,   cmake: '-DSDL2IMAGE_VENDORED=ON -G "Ninja Multi-Config"', msvc: 1 }
+        - { name: Windows (MSVC+CMake),         os: windows-latest, shell: sh,   cmake: '-DSDL2IMAGE_VENDORED=ON -GNinja', msvc: 1 }
         - { name: Windows (mingw32+autotools),  os: windows-latest, shell: 'msys2 {0}', msystem: mingw32, msys-env: mingw-w64-i686 }
         - { name: Windows (mingw64+CMake),      os: windows-latest, shell: 'msys2 {0}', msystem: mingw64, msys-env: mingw-w64-x86_64,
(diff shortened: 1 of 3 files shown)


madebr:
With travis-ci, using a recent CMake version was painful because it used an older ubuntu LTS release.
With github actions, we have the luxury of having an up-to-date cmake version.

As @smcv said, 3.16 would be a good compromise.



> I’m fine requiring a newer version of CMake for the new CMake support in the satellite libraries. I leave it to you guys to pick an appropriate version.

@slouken 
About CMake support of the main SDL library, what is the "official" position on this? Support at least CMake 3.0 until SDL3.0 gets released?

smcv:
> I'd like to support the various Linux and embedded use cases, but in general it's pretty easy for end users to upgrade CMake

... or continue to use Autotools, which was specifically designed to work on 1980s proprietary-Unix dinosaurs, so older Linux distributions are definitely fine.

slouken:
> > I'd like to support the various Linux and embedded use cases, but in general it's pretty easy for end users to upgrade CMake
> 
> ... or continue to use Autotools, which was specifically designed to work on 1980s proprietary-Unix dinosaurs, so older Linux distributions are definitely fine.

Exactly. @icculus, any issues with bumping main SDL's required CMake version to 3.16?

icculus:
(the oldest Ubuntu long-term-support release that is still supported is 18.04, which is the specific release that #4930 was concerned about, and its support ends in less than a year.)

smcv:
Being able to build on the newest Debian stable, Ubuntu LTS and RHEL seems like a useful cutoff point to have. Until recently, the oldest CMake in any of those was CMake 3.16 (Ubuntu 20.04).

Now that Ubuntu 22.04 has superseded 20.04, the next-oldest in conservative distributions is CMake 3.18 (Debian 11 and RHEL) - although note that our CI is still on Ubuntu 20.04 at the moment (but I think it has a newer CMake installed out-of-band).

slouken:
Okay, I'm fine with sticking with 3.16, or bumping up to 3.18 if that has changes that we'd like to use, and our CI supports it.

madebr:
I've settled on using CMake 3.16.
It is sufficiently new and has is distributed by (most) current Linux LTS releases.

Issue #257, opened 2022-05-23
macOS ImageIO corrupts .ico, .cur when using IMG_Load()


smcv:
Steps to reproduce:

* Build on macOS with the ImageIO backend
* Load `test/sample.ico`, `test/sample.cur` on macOS using `IMG_Load()`

Expected result: they load successfully and draw as a copy of the reference

Actual result: https://github.com/libsdl-org/SDL_image/issues/248#issuecomment-1133183528, https://github.com/libsdl-org/SDL_image/issues/248#issuecomment-1133188304

Workaround: Use `IMG_Load_RW()` or `IMG_LoadTyped_RW()` instead.

smcv:
As with #256, I'm only opening this so I have something to refer to in a commit: I don't think it's a release blocker.

I suspect that what is happening here is that the version of `IMG_Load()` in `IMG_ImageIO.m` thinks it can load these files, but its interpretation of them is not the same as ours.

Issue #256, opened 2022-05-23
tinyjpeg encoder corrupts images with width not divisible by 8


smcv:
Steps to reproduce:

* Configure with `-DSDL2IMAGE_BACKEND_STB=ON`, which uses `stb_image.h` for loading and `tiny_jpeg.h` for saving
* Configure with `-DSDL2IMAGE_SAMPLES=ON`
* Build
* `./showimage -quit -save times8.jpg test/times8.bmp -save sample.jpg test/sample.bmp`

where `times8.bmp` is this 24x40 bitmap converted to `.bmp`:

![times8](https://user-images.githubusercontent.com/1561141/169887595-62e48e67-f53e-4b99-8fd9-9dcbe0ebebda.png)

and `test/sample.bmp` is the 23x42 bitmap that I committed to this repo already.

Expected result: `times8.jpg` and `sample.jpg` look like their respective `.bmp` references, but with some quality loss.

Actual result: `times8.jpg` looks fine:

![times8](https://user-images.githubusercontent.com/1561141/169887431-063fbfc4-47e1-48ec-8326-f332400eb22b.jpg)

but `sample.jpg` has been mis-encoded:

![sample](https://user-images.githubusercontent.com/1561141/169887447-91a559f0-2f01-4cfe-aabc-35b676847a27.jpg)

Workaround: Use libjpeg, which gets this right.

smcv:
This is one of the issues involved in #248.

The most likely reason to save JPEGs from SDL is screenshots, in which case this would probably corrupt 1366x768 screenshots, but would be fine for every other commonly-used resolution.

sezero:
Issue doesn't seem to be reported to [mainstream](https://github.com/serge-rgb/TinyJPEG).

smcv:
I haven't done any real investigation on this; for now, I'm only opening this issue so that I have something to refer to in a commit that works around it.

slouken:
I can take a look at this. I'm pretty sure it's just a stride mismatch issue.

smcv:
It could also be to do with the 8x8 encoding blocks in the JPEG format.

slouken:
I've also filed an upstream PR: https://github.com/serge-rgb/TinyJPEG/pull/15

PR #258, opened 2022-05-23
Fix several tests, expand CI coverage

### LINKED ISSUE #248: Automated test failing on macOS

smcv:
* cmake: With vendored libjxl, only link libjxl_dec statically
    
    We only have a libjxl decoder, so we don't need the encoder half of the
    library, or its dependencies. This fixes a vendored build on Linux,
    which otherwise fails to configure because there is no vendored liblcms.

* cmake: If we're using a vendored libjxl, link it as C++
    
    Otherwise we'll fail to link with missing libstdc++ symbols.
    
    Thanks: Anonymous Maarten

* cmake: Correctly choose whether to enable or disable PNG saving
    
    This wasn't updated when the SAVE_PNG macro was renamed to
    SDL_IMAGE_SAVE_PNG.
    
    Fixes: 3af38ea "Fixed bug #235: add options to disable saving png/jpg images"

* test: Remove workaround for SDL_image#243
    
    The previous commit makes this unnecessary.

* test: Fix conversion to RGBA32
    
    We were early-returning due to a faulty check.
    
    Fixes: 59dff46 "Add a simple automated test"

* test: Fix confusing assertion messages
    
    The sense of these assertions is the reverse of what the message would
    suggest.
    
    Fixes: 59dff46 "Add a simple automated test"

* test: Use separate filenames for each saving test
    
    This lets us do a post-mortem comparison on them.

* test: Show the pixels we are comparing on test failure
    
    All our test images are relatively small, so this doesn't actually
    make the logs particularly huge.

* test: Increase tolerance of JXL decoding
    
    When running the tests with vendored libjxl, the results were visually
    indistinguishable from the reference image, but with more difference
    than our previous arbitrary limit (sum of squares of colour channels
    <= 100). Raise the arbitrary limit to 300.
    
    Fixes: 59dff46 "Add a simple automated test"

* test: Add some tolerance to AVIF decoding
    
    When running this test on macOS, the result was visually
    indistinguishable from the reference image, but not identical. Allow a
    small difference in its decoding (sum of squares of colour channel
    differences <= 300).
    
    Fixes: 59dff46 "Add a simple automated test"

* test: Use correct macro for PNG enabledness

* test: Account for Windows- and macOS-specific backends
    
    Some image formats can be loaded using OS facilities on Windows and
    macOS, even if we didn't enable them at configure time.

* workflows: Run automated test on macOS + Autotools
    
    With previous commits from this pull request, the test can now pass.
    
    This reverts commit ce31fd8779488390ce632b24368afaaabbf31350.

* workflows: Run the automated test for CMake Linux builds
    
    This exercises the "everything is vendored" use-case.
    
    We can't use dlopen() for the dependencies, because they won't be in
    the search path at runtime, so link them statically here.
    
    We also can't build the test on Windows (because MSYS' SDL packaging
    doesn't have libSDL_test) or on macOS with CMake (because libSDL_test
    isn't found), so skip those.

---

If successful, this will resolve #248, except for the parts that have been moved to #256 and #257.


smcv:
> Sigh, this needs more work: building against a vendored libjxl on Linux or macOS requires libstdc++ or equivalent.

In the short term, I can either disable libjxl for these CI builds, or use a non-vendored libjxl. Any preference?

slouken:
non-vendored libjxl for Linux is fine.

I'm surprised macOS is having issues though, I'm building that here using Xcode 13 with no issues. Are you using xcodebuild with the provided projects? If you do that you'll need to modify config.xcconfig, which disabled libjxl by default.

madebr:
> > Sigh, this needs more work: building against a vendored libjxl on Linux or macOS requires libstdc++ or equivalent.
> 
> In the short term, I can either disable libjxl for these CI builds, or use a non-vendored libjxl. Any preference?

I had a similar issue with SDL_ttf on Macos: linking failed because it couldn't find some c++ related symbols.
I fixed it by adding `enable_language(CXX)`.

https://github.com/madebr/SDL_ttf/blob/277a6c8e61c4973b1c09df61bfcad447ffbcab70/CMakeLists.txt#L170-L171

smcv:
> I fixed it by adding enable_language(CXX)

Thanks, trying that.

> Are you using xcodebuild with the provided projects?

No, I'm using Autotools and CMake on CI. I don't have access to a Mac, so the Github workflows are the only macOS builds I can do.

smcv:
> #257 should be fixed, I don't think you need this commit.

I did at the time I wrote the branch, but I think you're right that I don't any more!

.github/workflows/main.yml
```diff
@@ -19,8 +19,8 @@ jobs:
         - { name: Windows (mingw64+CMake),      os: windows-latest, shell: 'msys2 {0}', msystem: mingw64, msys-env: mingw-w64-x86_64,
             cmake: '-DSDL2IMAGE_BACKEND_STB=OFF -DSDL2IMAGE_BACKEND_WIC=OFF -DSDL2IMAGE_VENDORED=OFF -DSDL2IMAGE_AVIF=ON -G "Ninja Multi-Config"' }
         - { name: Linux (autotools),            os: ubuntu-20.04,   shell: sh }
-        - { name: Linux (CMake),                os: ubuntu-20.04,   shell: sh,   cmake: '-DSDL2IMAGE_VENDORED=ON -GNinja' }
-        - { name: 'Linux (CMake, static)',      os: ubuntu-20.04,   shell: sh,   cmake: '-DSDL2IMAGE_VENDORED=ON -DBUILD_SHARED_LIBS=OFF -GNinja' }
+        - { name: Linux (CMake),                os: ubuntu-20.04,   shell: sh,   cmake: '-DSDL2IMAGE_TESTS=ON -DSDL2IMAGE_VENDORED=ON -GNinja ' }
+        - { name: 'Linux (CMake, static)',      os: ubuntu-20.04,   shell: sh,   cmake: '-DSDL2IMAGE_TESTS=ON -DSDL2IMAGE_VENDORED=ON -DBUILD_SHARED_LIBS=OFF -GNinja' }
```
madebr, line 23: Can we move `-DSDL2IMAGE_TESTS=ON` to the workflow steps?
Are did you put them here because of the tests failing on Macos?

.github/workflows/main.yml
```diff
@@ -94,21 +94,61 @@ jobs:
         cmake -B build \
           -DBUILD_SHARED_LIBS=ON \
           -DCMAKE_BUILD_TYPE=Release \
+          -DSDL2IMAGE_DEPS_SHARED=OFF \
```
madebr, line 97: This is worth an issue, I think.
We should add the paths of the vendored dependencies + SDL2 to `PATH` on Windows (I think) , to `LD_LIBRARY_PATH` on Linux and `DYLD_LIBRARY_PATH` on Macos.

.github/workflows/main.yml
smcv: I put them here because there were enough exceptions and maybes that this seemed clearer than the way I'd previously been doing it, particularly while I was juggling all the various builds to get as many tests passing as I could.

Now that I've sorted out what does and doesn't work, they could move back to the `cmake` invocation as "only if Linux", I think.

smcv: > did you put them here because of the tests failing on Macos?

No, they're here because the tests don't compile on macOS with CMake (it looks like Autotools can find `-lSDL_test`, but CMake can't, for whatever reason).

.github/workflows/main.yml
smcv: The paths where the libraries get compiled are an implementation detail of the vendored dependencies, and different for each vendored dependency, so that's going to be quite a long search path. Setting the right environment variable is going to require platform-specific knowledge: libtool and CMake both know how to do this for ordinary linking with `-lfoo`, but not for dynamic loading via `dlopen()`.

.github/workflows/main.yml
madebr: I think it's because we want to link to `SDL2_test` as `-lSDL2_test`, but with Brew libraries are not available in the usual locations.
Perhaps we need to do things like:
```cmake
find_library(SDL2_TEST_LIBRARY
     NAMES SDL2_test
)
if(SDL2_TEST_LIBRARY AND NOT TARGET SDL2::SDL2test)
    add_library(SDL2::SDL2test UNKNOWN IMPORTED)
    set_target_properties(SDL2::SDL2test PROPERTIES
        IMPORTED_LOCATION "${SDL2_TEST_LIBRARY}"
    )
endif()
```
+ throw an error when SDL2_test is not found.

smcv: I believe this PR is an improvement with no regressions: it fixes some bugs (not just in the test but also in the functionally necessary code), and the number of platforms tested goes up. Previously we weren't running tests for macOS at all, now we're testing macOS+Autotools which is better than nothing; and previously we were running tests with Linux+Autotools, but now we're also testing Linux+CMake.

Also the branch is getting quite long. So if possible I'd prefer to get the changes-so-far merged before doing the necessary trial-and-error to find this library on a remote system I can only access via CI.

test/main.c
```diff
@@ -530,6 +528,96 @@ ConvertToRgba32(SDL_Surface **surface_p)
...
+    const unsigned char *p;
+    size_t w, h, pitch;
+    size_t i, j;
+
+    fprintf(stderr, "%s:\n", filename);
+
+    if (surface->format->palette) {
+        size_t n = surface->format->palette->ncolors;
+
+        if (n < 0)
```
sezero, line 544: this check can be removed here - `n` can't be negative.

.github/workflows/main.yml
smcv: I'm opening an issue.

test/main.c
smcv: It should be a check on `surface->format->palette->ncolors` before we assign it to `n`, ideally.

.github/workflows/main.yml
smcv: I'll try this:

> Can we move -DSDL2IMAGE_TESTS=ON to the workflow steps?

but I'm not intending to include running the test on macOS+CMake in the scope of this branch.

madebr: It won't build on Macos due to SDL2_test not being found. So I wouldn't bother.

test/main.c
sezero: Let's fix it that way, then

.github/workflows/main.yml
smcv: #259

.github/workflows/main.yml
smcv: SDL_test not found on macOS → #260

.github/workflows/main.yml
```diff
@@ -91,24 +91,78 @@ jobs:
...
     - name: Build
       if: "matrix.platform.cmake"
       run: cmake --build build/ --config Release --verbose
+    - name: Run build-time tests
+      if: "runner.os == 'Linux' && matrix.platform.cmake"
+      run: |
+        set -eu
+
+        # Expect to support all formats except AVIF in these builds
+        export SDL_IMAGE_TEST_REQUIRE_LOAD_AVIF=0
```
madebr, line 126: For a future pr, I suppose it is ok to move these environment variables to CMake?
[`ENVIRONMENT` property on tests](https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html)

smcv: No, not really: these are here as a way for the CI to demand that PNG is supported, even if the build system has become confused and disabled PNG due to failed auto-detection or a build system bug.

Specifically, I wanted these for Debian, to detect regressions like the one introduced in 3af38ea and fixed in d25c532 (which accidentally disabled PNG saving in the packages I maintain in Debian experimental). I want to be able to set `SDL_IMAGE_TEST_REQUIRE_LOAD_PNG=1` in the Debian CI, and have that make the test fail if it can't load PNGs.

smcv: If the CMake build system was responsible for setting these variables, then a bug in the build system that accidentally disabled PNG wouldn't be detected, because the test (which would be told by the build system that PNG was disabled) would be consistent with the implementation (which genuinely did disable PNG).

madebr: Understood, as a packager it's understandable to not trust the build system to do *the right thing*.
(we should document this)

smcv:
3cf3fe2: adjusted to avoid conflicts with #261, no functional changes.

PR #264, opened 2022-05-25
Fix some formats on big-endian machines

smcv:
* jpg: Fix channel order of saved JPEGs on big-endian platforms
    
    Both of the JPEG libraries we use (jpeglib or tinyjpeg) expect to see
    the red channel in the first byte of the buffer, regardless of whether
    the first byte is considered to be the most-significant or
    least-significant in native endianness.

* tiny_jpeg: Don't byteswap 16-bit words if already big-endian
    
    Previously, this assumed that native endianness was always little-endian,
    which is true for x86 and most other architectures, but untrue for
    some (mostly older) unusual architectures.
    
    Forwarded: https://github.com/serge-rgb/TinyJPEG/pull/16

* qoi: Fix channel order on big-endian platforms
    
    qoi.h always decodes the red channel of the first pixel into the first
    byte of the buffer, regardless of whether interpreting the buffer as
    a sequence of 32-bit words would consider that to be the most or least
    significant byte of the first word. In SDL terminology that's
    SDL_PIXELFORMAT_RGBA32.

* svg: Fix channel order on big-endian platforms
    
    nanosvg always decodes the red channel of the first pixel into the first
    byte of the buffer, regardless of whether interpreting the buffer as
    a sequence of 32-bit words would consider that to be the most or least
    significant byte of the first word. In SDL terminology that's
    SDL_PIXELFORMAT_RGBA32.

### DIFF
--- a/IMG_jpg.c
+++ b/IMG_jpg.c
@@ -492,11 +492,9 @@ static void jpeg_SDL_RW_dest(j_compress_ptr cinfo, SDL_RWops *ctx)
 
 static int IMG_SaveJPG_RW_jpeglib(SDL_Surface *surface, SDL_RWops *dst, int freedst, int quality)
 {
-#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+    /* The JPEG library reads bytes in R,G,B order, so this is the right
+     * encoding for either endianness */
     static const Uint32 jpg_format = SDL_PIXELFORMAT_RGB24;
-#else
-    static const Uint32 jpg_format = SDL_PIXELFORMAT_BGR24;
-#endif
     struct jpeg_compress_struct cinfo;
     struct my_error_mgr jerr;
     JSAMPROW row_pointer[1];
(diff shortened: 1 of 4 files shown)


smcv:
Context for this is that I tried uploading a git snapshot of SDL_image to Debian experimental, where it failed build-time tests. I think this has been wrong forever, we just didn't notice until now, because nobody actually used SDL_image on s390x (or at least, not with the affected formats).

We don't normally enable AVIF and JXL in Debian, and we normally use jpeglib rather than tinyjpeg, but I tried all of those while testing this on a s390x developer machine.

smcv:
This might still be wrong for the formats that are not exercised by `test/main.c`: LBM, PCX, TGA, XV. I'm not intending to test or fix those as part of this PR.

If someone can convert `test/sample.bmp` into those formats, or provide small test files in those formats under the Zlib license (together with a known-good `.bmp` reference equivalent), then we can add test coverage for those later (and I can try them on s390x if asked).

sezero:
Thank you.

PR #265, opened 2022-05-25
test: Enable automated test for TGA format

smcv:
* test: Make sample files non-executable

* test: Enable automated test for TGA format
    
    SDL_image currently has no way to recognise this format from content,
    only from filename. Other than that, it's the same as any other format.


smcv:
I confirm that this also works on Debian s390x (so our handling of TGA does not have endianness bugs like #264, at least for the particular format used in the sample image that @slouken committed).

sezero:
Thanks.

Issue #262, opened 2022-05-24
IMG_gif.c uses SDL_HasColorKey, which is SDL2 2.0.9+


madebr:
It's used here:
https://github.com/libsdl-org/SDL_image/blob/fe2c0c49cd429468870f2be0183a0294ad8fd343/IMG_gif.c#L150

Found while building SDL_image on MSVC:
https://github.com/madebr/SDL_image/runs/6582119390?check_suite_focus=true

Documentation of [SDL_HasColorKey](https://wiki.libsdl.org/SDL_HasColorKey)

Perhaps we need to bump the minimum required SDL version to 2.0.9?
If this is ok, please create a github release (including binaries) for sdl 2.0.9.

sezero:
Pushed https://github.com/libsdl-org/SDL_image/commit/aa690a6de7e33b5456737b98abd512bb120f2f5e to bump required SDL version to 2.0.9

Keeping this open until @slouken publishes a github release for it.

sezero:
A 2.0.9 github release is created - closing.

madebr:
Thanks, though it has an issue:
the 2.0.9 release contains SDL2-devel-2.0.8-VC.zip, instead of SDL2-devel-2.0.9-VC.zip

https://github.com/libsdl-org/SDL/releases/tag/release-2.0.9

slouken:
Fixed, thanks!

sezero:
Actually, wouldn't it be better that we use `--no-undefined` linker flag
if available?
https://github.com/libsdl-org/SDL/blob/main/configure.ac#L250
https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L608

I can copy it to all satellite projects' autotools if accepted, and @madebr
can handle cmake side.

slouken:
Yep, that sounds good.

sezero:
Pushed autofoo patches to SDL_mixer, SDL_image, and SDL_ttf:
https://github.com/libsdl-org/SDL_mixer/commit/8bdfdc3a88cf72151ae4639098ab30dc0ddba8ae
https://github.com/libsdl-org/SDL_ttf/commit/274ad17e98d07dc6b4ba5ddb7bcb782767832297
https://github.com/libsdl-org/SDL_image/commit/a4c08c025df3115fc8216799734cdd15b25aced8

@madebr: For cmake, here is a fair example we have been using in libxmp:
[...]

PR #269, opened 2022-05-25
Don't allow undefined symbols, when buildind a shared SDL2_image

madebr:
Suggested in https://github.com/libsdl-org/SDL_image/issues/262#issuecomment-1136543597

@sezero


cmake/PrivateSdlFunctions.cmake
```diff
@@ -218,3 +218,26 @@ macro(sdl_check_project_in_subfolder relative_subfolder name)
             "Run the download script in the external folder, or re-configure with -DSDL2IMAGE_VENDORED=OFF to use system packages.")
     endif()
 endmacro()
+
+macro(sdl_check_linker_flag flag var)
+    # FIXME: Use CheckLinkerFlag module once cmake minimum version >= 3.18
+    include(CMakePushCheckState)
+    include(CheckCSourceCompiles)
+    cmake_push_check_state(RESET)
+    set(CMAKE_REQUIRED_LINK_OPTIONS "-Wl,--no-undefined")
```
sezero, line 227: I'm OK with this to get the functionality in. But: Can't this helper macro get an additional arg instead of hard-coding `-Wl,--no-undefined` here again? (sorry, not good at cmake.)

madebr: I'm afraid not.
The `check_c_source_compiles` is itself a wrapper around [`try_compile`](https://cmake.org/cmake/help/latest/command/try_compile.html#command:try_compile).
See https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html

`try_compile` accepts a `LINK_OPTIONS` argument, but it also requires additional arguments that would complicate the use here, such as bindir, and the need of saving the test source to disk ourselves.
Using the convenience function `check_c_source_compiles` also caches the result.

sezero: OK then, taking this as is.

PR #278, opened 2022-06-06
Fix memory leak in qoi decoder

niclasr:
Fix memory leak in qoi decoder.
SDL_memcpy copies the data, it does not
move the pointer, so free it after use
just like it is done if surface creation
fails.


sezero:
This looks correct to me: @slouken?

Although, reading of whole image to memory and duplicate allocation of the result
are very inefficient. Maybe we should customize qoi.h for our own use.

slouken:
Yep, we want to avoid the copy. Thanks!

PR #281, opened 2022-06-17
cmake: Fix static linking to dependencies with "-" in library name

smcv:
If we gain a dependency on a library with "-" in its name, we'd parse
the list of dependencies incorrectly; fix that. Equivalent to
https://github.com/libsdl-org/SDL/pull/5789 in SDL.


smcv:
mixer, net and ttf already have this fixed.

PR #291, opened 2022-07-13
Fix to find dynamic libraries

madebr:
After #290, I looked again at CMake's code to find the location of dynamic libraries.
CMake finds libraries in a different way, but should result in the same final library to be dynamically loaded.

CMake does this in 2 steps:
1. Use [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) to find a dependency.
2. Use the library location from step 1 to create the dynamic library name:
  - I noticed that in install prefixes, dynamic links are used to point between files.
    e.g. on my system, there are 2 symblinks to `/usr/lib64/libjxl.so.0.6.1`
    ```
    lrwxrwxrwx. 1 root root      13 20 jan 15:05 /usr/lib64/libjxl.so -> libjxl.so.0.6
    lrwxrwxrwx. 1 root root      15 20 jan 15:05 /usr/lib64/libjxl.so.0.6 -> libjxl.so.0.6.1
    -rwxr-xr-x. 1 root root 2916224 20 jan 15:06 /usr/lib64/libjxl.so.0.6.1
    ```
    ```
    - CMake finds all links in the same folder that point to `libjxl.so.0.6.1`, sorts these by name and selects the first one.
      This excludes the dynamic libraries ending with `.so` and `.dylib`.
    - This should always result in the one with the lowest string length to be chosen.

If the above fails, then simply use the library from step 1.


cmake/PrivateSdlFunctions.cmake
```diff
@@ -169,38 +169,37 @@ function(target_get_dynamic_library DEST TARGET)
...
-                    set(similar_files "${propvalue}")
+                    set(similar_files "${library_path}")
                     foreach(subfile ${subfiles})
                         if (IS_SYMLINK "${subfile}")
                             read_absolute_symlink(subfile_target "${subfile}")
-                            if (subfile_target STREQUAL propvalue)
+                            while (IS_SYMLINK "${subfile_target}")
+                                read_absolute_symlink(subfile_target "${subfile_target}")
+                            endwhile()
+                            if (subfile_target STREQUAL library_path AND NOT "${subfile}" MATCHES ".*(dylib|so)$")
```
slouken, line 195: And dll?

slouken:
This generally seems fine to me. Do you handle the case where these files all exist?
```
lrwxrwxrwx  1 Sam Lantinga None 14 Jul 13 07:23 libjxl.so -> libjxl.so.0.10
lrwxrwxrwx  1 Sam Lantinga None 16 Jul 12 11:50 libjxl.so.0.10 -> libjxl.so.0.10.1
-rw-rw-r--+ 1 Sam Lantinga None  0 Jul 12 11:50 libjxl.so.0.10.1
lrwxrwxrwx  1 Sam Lantinga None 15 Jul 12 08:40 libjxl.so.0.6 -> libjxl.so.0.6.1
-rw-rw-r--+ 1 Sam Lantinga None  0 Jul 12 08:39 libjxl.so.0.6.1
lrwxrwxrwx  1 Sam Lantinga None 15 Jul 12 10:18 libjxl.so.0.7 -> libjxl.so.0.7.1
-rw-rw-r--+ 1 Sam Lantinga None  0 Jul 12 10:18 libjxl.so.0.7.1
```

cmake/PrivateSdlFunctions.cmake
madebr: Dll's are handled by a different code path.
The `find_package` will find the import library and `dumpbin` or `dlltool` will extract the target dll.
No manual finding is required.

madebr:
> This generally seems fine to me. Do you handle the case where these files all exist?
> 
> ```
> lrwxrwxrwx  1 Sam Lantinga None 14 Jul 13 07:23 libjxl.so -> libjxl.so.0.10
> lrwxrwxrwx  1 Sam Lantinga None 16 Jul 12 11:50 libjxl.so.0.10 -> libjxl.so.0.10.1
> -rw-rw-r--+ 1 Sam Lantinga None  0 Jul 12 11:50 libjxl.so.0.10.1
> lrwxrwxrwx  1 Sam Lantinga None 15 Jul 12 08:40 libjxl.so.0.6 -> libjxl.so.0.6.1
> -rw-rw-r--+ 1 Sam Lantinga None  0 Jul 12 08:39 libjxl.so.0.6.1
> lrwxrwxrwx  1 Sam Lantinga None 15 Jul 12 10:18 libjxl.so.0.7 -> libjxl.so.0.7.1
> -rw-rw-r--+ 1 Sam Lantinga None  0 Jul 12 10:18 libjxl.so.0.7.1
> ```

In this case, our `find_package` will pick up `libjxl.so`.
The search algorithm will detect that `libjxl.so.0.10` and `libjxl.so.0.10.1` also point to the same file (or are the target file).
Since `libjxl.so.0.10` is the shortest, that one will be picked.

The other libraries are not even considered.

slouken:
Perfect. Can you apply the same fixes to SDL and SDL_mixer?

PR #309, opened 2022-10-11
Fixes for android prefab script + fix for vendored PNG

madebr:
- Fix using a shared vendored libpng, (candidate for the next patch release)
- Update libwebp to include the patch renaming cpufeatures target (it clashes with jpeg from libjxl)
- Give every 3rd party library a known supported name: e.g. make libpng available as `SDL2_image::libpng`.
- The current android prefab script used stb for loading images, and no external libraries.
 This update includes all vendored libraries, except libjxl.

Renaming the 3rd party vendored libraries to a known name is currently useful for Android.
To make sure the android build systems includes `libpng16.so` (and its dependencies) to the `.apk`, you can now add the following to the cmake script:
```cmake
if(ANDROID AND TARGET SDL2_image::external_libpng)
    target_link_libraries(mygame PRIVATE SDL2_image::external_libpng)
endif()
```
This is not 100% perfect because this introduces an explicit link, but works for now.
I've created https://github.com/google/prefab/issues/159 and https://github.com/google/prefab/issues/160 asking how to handle this.


madebr:
FYI, I've tested these changes in my [sdlcross](https://github.com/madebr/sdlcross/blob/master/CMakeLists.txt) try-out project.

Issue #314, opened 2022-10-31
webp image animation support


zeromake:
webp image is support animation. 
Hope this feature can be added.

Below is my own trial code:

**IMG_webp.c**

```c
#include <webp/decode.h>
#include <webp/demux.h>

IMG_Animation *IMG_LoadWEBPAnimation_RW(SDL_RWops *src)
{
    Sint64 start;
    const char *error = NULL;
    Uint32 Rmask;
    Uint32 Gmask;
    Uint32 Bmask;
    Uint32 Amask;
    WebPBitstreamFeatures features;
    struct WebPDemuxer* dmuxer = NULL;
    IMG_Animation *anim = NULL;
    int raw_data_size;
    uint8_t *raw_data = NULL;
    int r;
    uint8_t *ret;

    if ( !src ) {
        /* The error message has been set in SDL_RWFromFile */
        return NULL;
    }

    start = SDL_RWtell(src);

    if ( (IMG_Init(IMG_INIT_WEBP) & IMG_INIT_WEBP) == 0 ) {
        goto error;
    }

    raw_data_size = -1;
    if ( !webp_getinfo( src, &raw_data_size ) ) {
        error = "Invalid WEBP Animation";
        goto error;
    }

    raw_data = (uint8_t*) SDL_malloc( raw_data_size );
    if ( raw_data == NULL ) {
        error = "Failed to allocate enough buffer for WEBP Animation";
        goto error;
    }

    r = (int)SDL_RWread(src, raw_data, 1, raw_data_size );
    if ( r != raw_data_size ) {
        error = "Failed to read WEBP Animation";
        goto error;
    }

    if ( lib.WebPGetFeaturesInternal( raw_data, raw_data_size, &features, WEBP_DECODER_ABI_VERSION ) != VP8_STATUS_OK ) {
        error = "WebPGetFeatures has failed";
        goto error;
    }

    /* Check if it's ok !*/
#if SDL_BYTEORDER == SDL_LIL_ENDIAN
    Rmask = 0x000000FF;
    Gmask = 0x0000FF00;
    Bmask = 0x00FF0000;
    Amask = (features.has_alpha) ? 0xFF000000 : 0;
#else
    {
        int s = (features.has_alpha) ? 0 : 8;
        Rmask = 0xFF000000 >> s;
        Gmask = 0x00FF0000 >> s;
        Bmask = 0x0000FF00 >> s;
        Amask = 0x000000FF >> s;
    }
#endif

[...]

PR #315, opened 2022-11-01
cmake: animation support links to functions from libwebpdemux

madebr:
Link to libwebpdemux as well, for [pi](https://github.com/libsdl-org/SDL_image/issues/314)


sezero:
We should handle `LOAD_WEBP_DYNAMIC` case: maybe add a new `LOAD_WEBPDEMUX_DYNAMIC` and adjust IMG_webp.c for it?

The autotools side should be touched accordingly after this, too.

madebr:
We need 2 handles then, one for `libwebp.so.7` (as it is used now) and one for `libwebpdemux.so.2`? Right?

sezero:
> We need 2 handles then, one for `libwebp.so.7` (as it is used now) and one for `libwebpdemux.so.2`? Right?

Yes, that's what I think.

Then again, libwebpdemux.so _should_ be linked to libwebp.so, so: Can we just simply change the handle from libwebp.so to libwebpdemux.so for simplicity? (No, haven't tried myself yet.)

sezero:
With current main (i.e. make sure to see that commit 1af8a539a is applied), the following autotools patch works for me (generated file changes not included.)

```diff
diff --git a/configure.ac b/configure.ac
index 3358d43..9b0be17 100644
--- a/configure.ac
+++ b/configure.ac
@@ -273,8 +273,6 @@ AC_ARG_ENABLE([xv], [AS_HELP_STRING([--enable-xv], [support loading XV images [d
  [], [enable_xv=yes])
 AC_ARG_ENABLE([webp], [AS_HELP_STRING([--enable-webp], [support loading WEBP images [default=yes]])],
  [], [enable_webp=yes])
-AC_ARG_ENABLE([webpdec], [AS_HELP_STRING([--enable-webpdec], [support loading WEBP images via libwebpdecoder instead of libwebp [default=no]])],
- [], [enable_webpdec=no])
 AC_ARG_ENABLE([webp-shared], [AS_HELP_STRING([--enable-webp-shared], [dynamically load WEBP support [default=yes]])],
  [], [enable_webp_shared=yes])
 AC_ARG_ENABLE([qoi], [AS_HELP_STRING([--enable-qoi], [support loading QOI images [default=yes]])],
@@ -534,25 +532,19 @@ if test x$enable_tif = xyes -a x$enable_imageio != xyes; then
 fi
 AC_SUBST([LOAD_TIF], $load_tif)
 
-load_webp=0
-if test x$enable_webpdec = xyes; then
-  webplib=webpdecoder
-else
-  webplib=webp
-fi
 if test x$enable_webp = xyes; then
-    PKG_CHECK_MODULES([LIBWEBP], [lib$webplib], [dnl
+    PKG_CHECK_MODULES([LIBWEBPDEMUX], [libwebpdemux], [dnl
         have_webp_hdr=yes
         have_webp_lib=yes
         have_webp_pc=yes
       ], [dnl
-        AC_CHECK_HEADER([webp/decode.h], [
+        AC_CHECK_HEADER([webp/demux.h], [
             have_webp_hdr=yes
-            LIBWEBP_CFLAGS=""
+            LIBWEBPDEMUX_CFLAGS=""
         ])
-        AC_CHECK_LIB([$webplib], [WebPGetDecoderVersion], [
+        AC_CHECK_LIB([webpdemux], [WebPDemuxGetFrame], [
             have_webp_lib=yes
-            LIBWEBP_LIBS="-l$webplib"
+            LIBWEBPDEMUX_LIBS="-lwebpdemux"
         ], [], [-lm])
       ])
     if test x$have_webp_hdr = xyes -a x$have_webp_lib = xyes; then
@@ -561,15 +553,15 @@ if test x$enable_webp = xyes; then
 
         case "$host" in
             *-*-darwin*)
-                webp_lib=[`find_lib lib$webplib.dylib`]
+                webp_lib=[`find_lib libwebpdemux.dylib`]
                 ;;
             *-*-cygwin* | *-*-mingw*)
-                webp_lib=[`find_lib "lib$webplib-*.dll"`]
+                webp_lib=[`find_lib "libwebpdemux-*.dll"`]
                 ;;
             *)
-                webp_lib=[`find_lib "lib$webplib[0-9]*.so.*"`]
+                webp_lib=[`find_lib "libwebpdemux[0-9]*.so.*"`]
                 if test x$webp_lib = x; then
-                    webp_lib=[`find_lib "lib$webplib.so.*"`]
+                    webp_lib=[`find_lib "libwebpdemux.so.*"`]
                 fi
                 ;;
         esac
@@ -663,11 +655,11 @@ if test x$enable_webp = xyes -a x$have_webp_hdr = xyes -a x$have_webp_lib = xyes
         echo "-- dynamic libwebp -> $webp_lib"
         AC_DEFINE_UNQUOTED(LOAD_WEBP_DYNAMIC, "$webp_lib")
     else
-        IMG_LIBS="$LIBWEBP_LIBS $IMG_LIBS"
+        IMG_LIBS="$LIBWEBPDEMUX_LIBS $IMG_LIBS"
         if test x$have_webp_pc = xyes; then
-            PC_REQUIRES="lib$webplib $PC_REQUIRES"
+            PC_REQUIRES="libwebpdemux $PC_REQUIRES"
         else
-            PC_LIBS="$LIBWEBP_LIBS $PC_LIBS"
+            PC_LIBS="$LIBWEBPDEMUX_LIBS $PC_LIBS"
         fi
     fi
 fi
```

madebr:
It's okay to push separately. autotools and cmake are seaparate systems and won't cause merge conflicts.

madebr:
I don't think using one library works on Windows.
Compiling the following code:
```
#include <stdio.h>
#include <windows.h>

int main() {
    void* h = LoadLibraryA("libwebpdemux-2.dll");
    printf("h=%p\n", h);
    void*r;
    r = GetProcAddress(h, "WebPDecodeRGBInto");
    printf("WebPDecodeRGBInto -> %p\n", r);
    r = GetProcAddress(h, "WebPDemuxGetFrame");
    printf("WebPDemuxGetFrame -> %p\n", r);
    FreeLibrary(h);
    return 0;
}
```

and running under wine gives:
```
#include <stdio.h>
#include <windows.h>

int main() {
    void* h = LoadLibraryA("libwebpdemux-2.dll");
    printf("h=%p\n", h);
    void*r;
    r = GetProcAddress(h, "WebPDecodeRGBInto");
    printf("WebPDecodeRGBInto -> %p\n", r);
    r = GetProcAddress(h, "WebPDemuxGetFrame");
    printf("WebPDemuxGetFrame -> %p\n", r);
    FreeLibrary(h);
    return 0;
}
```

I think we really need to use 2 library handles.

sezero:
> I don't think using one library works on Windows. Compiling the following code:
>
> I think we really need to use 2 library handles.

Well, hell.. `GetProcAddress` is tied to the module handle, so you are correct.

Can you handle autofoo side too?

madebr:
I'll try :)

IMG_webp.c
```diff
@@ -62,37 +63,47 @@ static struct {
 } lib;
 
 #ifdef LOAD_WEBP_DYNAMIC
```
sezero, line 65: this should be changed to `#if defined(LOAD_WEBP_DYNAMIC) && defined(LOAD_WEBPDEMUX_DYNAMIC)`

IMG_webp.c
```diff
@@ -62,37 +63,47 @@ static struct {
...
     if (lib.FUNC == NULL) { IMG_SetError("Missing webp.framework"); return -1; }
+#define FUNCTION_LOADER_LIBWEBPDEMUX(FUNC, SIG) \
+    lib.FUNC = FUNC; \
+    if (lib.FUNC == NULL) { IMG_SetError("Missing webpdemux.framework"); return -1; }
 #endif
 
 int IMG_InitWEBP()
 {
     if (lib.loaded == 0) {
 #ifdef LOAD_WEBP_DYNAMIC
```
sezero, line 84: same as above

IMG_webp.c
```diff
@@ -105,7 +116,8 @@ void IMG_QuitWEBP()
     }
     if (lib.loaded == 1) {
 #ifdef LOAD_WEBP_DYNAMIC
```
sezero, line 118: same as above

IMG_webp.c
madebr: force pushed these changes

sezero:
This looks OK to me.

PR #316, opened 2022-11-23
Fix samples install path

cgohlke:


madebr (changes requested):
Yup, I'm guilty of copy pasting :)

madebr (approved):

PR #318, opened 2022-12-02
Replace SDL_CreateRGBSurface by SDL_CreateRGBSurfaceWithFormat

1bsyl:
Replace SDL_CreateRGBSurface by SDL_CreateRGBSurfaceWithFormat
so that it will be easier to move to SDL3

### DIFF
--- a/IMG.c
+++ b/IMG.c
@@ -185,7 +185,7 @@ SDL_Surface *IMG_Load(const char *file)
 
     data = emscripten_get_preloaded_image_data(file, &w, &h);
     if (data != NULL) {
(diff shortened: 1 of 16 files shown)


sezero (approved):
This looks OK to me. @slouken?

madebr:
We need to have a SDL2_image branch first.

And for the time being, build+install SDL3 before building SDL3_image on CI (see https://github.com/libsdl-org/SDL/issues/6667).

sezero:
> We need to have a SDL2_image branch first.

This is not an SDL3 patch: This builds against SDL2 just fine.

slouken:
Yup!

PR #331, opened 2023-01-29
Fix loading of 4-bit .ico files

SpaceManiac:
`biBitCount` is used on line 761 to infer the default palette size if `biClrUsed` is zero, but is set to 8 here for no apparent reason. Of course a 16-color image doesn't have a 256-color palette. It seems the bug has existed forever.

Example 4-bit .ico file that fails to load before but loads with this patch:

PR'd to `release-2.6.x` branch out of optimism that this fix could make it into a future 2.x release; I don't know your backport policy. Can rebase if needed.

### DIFF
--- a/IMG_bmp.c
+++ b/IMG_bmp.c
@@ -240,7 +240,6 @@ LoadICOCUR_RW(SDL_RWops * src, int type, int freesrc)
         case 1:
         case 4:
             ExpandBMP = biBitCount;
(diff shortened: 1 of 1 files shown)


slouken:
Please rebase it to main, and we'll backport to the release branch.

Thanks!

slouken:
Merged, thanks!

PR #333, opened 2023-01-31
[PoC] Use GitHub artifacts for ci

madebr:
Since https://github.com/libsdl-org/SDL/pull/7113, libsdl-org/SDL provides binary SDL3 artifacts.
This PR is a proof of concept to use these instead of (currently) non-existing SDL3 system packages.

The issues with this approach are:
- I'm not 100% sure the SDL3 artifact comes from the main branch (it does not filter out pr's etc)
- This uses the GitHub REST api which has usage constraints. If a lot of network traffic happens, action can be temporarily blocked.
- The GitHub action should be moved to some central place because this is also useful for the other satellite libraries.

Ideally, we provide a rolling SDL3 release somewhere on a dedicated repo. e.g. at `libsdl-org/setup-sdl`


slouken:
@icculus, instead of storing binary artifacts on GitHub, can we add a workflow to store them on libsdl.org?

slouken:
Maybe we can kick off a job via URL on the website on a successful CI run to sync and build there?

icculus:
There is already a webhook we can use to notify our server of successful CI builds, fwiw, but I need to reread this PR like three more times to understand what we're doing here.  :)

PR #352, opened 2023-03-26
Don't add -Wl,--no-undefined for clang, windows 10

AlpRabbit:
Since clang on Windows 10 doesn't have `-Wl,--no-undefined` option, disabling it. [Same thing is done in SDL](https://github.com/libsdl-org/SDL/commit/34ecd71e80b651c9840f60d91be60e30305fdc6f).


sezero:
The real problem is why is the cmake test not catching that that linker option isn't supported, not just here but also in SDL itself?

AlpRabbit:
I made a mistake I will open it again.

madebr (approved):

Issue #369, opened 2023-06-18
Move all sources into subdirectories


madebr:
- `SDL_Image` sources should go into `/src`, with eventual subdirectories
- showanim.c and showimage.c go into `/examples`.

slouken:
This is fine, feel free to make this change for SDL3. For SDL2 it's more complicated because of the automake configuration.

madebr:
This has been done in #370 and #371

PR #370, opened 2023-06-20
[SDL3] Move sources and documentation in subfolders

madebr:
Addresses the SDL3 branch for #369

Untested changes include:
- Android.mk
- VC and VC-UWP
- xcode project
- `build-scripts/test-driver` (I don't know how to run this file)


sezero:
When we did the same in SDL_mixer, we moved the public header under include directory: should we do the same here? (Same for the SDL2 version of the patch.)

madebr:
> When we did the same in SDL_mixer, we moved the public header under include directory: should we do the same here? (Same for the SDL2 version of the patch.)

I did this in the SDL2_image pr. SDL3_image does not need this.

PR #368, opened 2023-06-17
ci: try my experimental setup-sdl action

madebr:
This is a first step in fixing https://github.com/libsdl-org/SDL/issues/6667 and https://github.com/libsdl-org/SDL/issues/7219.
I've created a GitHub action at https://github.com/madebr/setup-sdl, that can be included in a workflow yaml using:
```yml
    - name: Set up SDL
      id: sdl
      uses: madebr/setup-sdl@master
      with:
        version: 3-head
```
The arguments are listed in [`action.yml`](https://github.com/madebr/setup-sdl/blob/master/action.yml#L3-L11).
- 2.x.y: use a particular SDL2 release
- 2-any: use any SDL2 release
- 2-latest: use the latest SDL2 release
- 2-head: use the SDL2 branch from the SDL repo
- 3-head: use the main branch from the SDL repo
- git hash: use a particular git hash from the SDL repo

It simply builds SDL with everything defaulted, and no caching.
Also, no system dependencies are downloaded or installed.
This currently only works when using the default toolchains of Linux, Macos and Windows runners.

~For some reason [it fails with Visual Studio](https://github.com/madebr/issues/actions/runs/5295593258/jobs/9586118132).~

PR #373, opened 2023-07-01
use libsdl-org/setup-sdl for providing SDL2

madebr:
This is the SDL2 equivalent of #368

For some reason autotools fails to build the examples.
I don't understand the reason, because it locally works in a docker container by either setting `SDL2_CONFIG` or `PKG_CONFIG_PATH`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,24 +15,28 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows (MSVC+CMake),         os: windows-latest, shell: sh,   cmake: '-DSDL2IMAGE_VENDORED=ON -GNinja', msvc: 1, shared: 1, static: 0 }
+        - { name: Windows (MSVC+CMake),         os: windows-latest, shell: sh, cmake: '-GNinja', msvc: 1, shared: 1, static: 0, vendored: 1 }
         - { name: Windows (mingw32+autotools),  os: windows-latest, shell: 'msys2 {0}', msystem: mingw32, msys-env: mingw-w64-i686, shared: 1, static: 1 }
         - { name: Windows (mingw64+CMake),      os: windows-latest, shell: 'msys2 {0}', msystem: mingw64, msys-env: mingw-w64-x86_64, shared: 1, static: 0,
-            cmake: '-DSDL2IMAGE_BACKEND_STB=OFF -DSDL2IMAGE_BACKEND_WIC=OFF -DSDL2IMAGE_VENDORED=OFF -DSDL2IMAGE_AVIF=ON -G "Ninja Multi-Config"' }
+            cmake: '-DSDL2IMAGE_BACKEND_STB=OFF -DSDL2IMAGE_BACKEND_WIC=OFF -G "Ninja Multi-Config"' }
         - { name: Linux (autotools),            os: ubuntu-latest,  shell: sh, shared: 1, static: 1}
-        - { name: Linux (CMake),                os: ubuntu-latest,  shell: sh,  cmake: '-DSDL2IMAGE_VENDORED=ON -GNinja', shared: 1, static: 0 }
-        - { name: 'Linux (CMake, static)',      os: ubuntu-latest,  shell: sh,   cmake: '-DSDL2IMAGE_VENDORED=ON -DBUILD_SHARED_LIBS=OFF -GNinja', shared: 0, static: 1 }
+        - { name: Linux (CMake),                os: ubuntu-latest,  shell: sh, cmake: '-GNinja', shared: 1, static: 0, vendored: 1 }
+        - { name: 'Linux (CMake, static)',      os: ubuntu-latest,  shell: sh, cmake: '-DBUILD_SHARED_LIBS=OFF -GNinja', shared: 0, static: 1, vendored: 1 }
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -14941,7 +14941,7 @@ fi
                 jxl_lib=`find_lib libjxl.dylib`
                 ;;
             *-*-cygwin* | *-*-mingw*)
-                jxl_lib=`find_lib "libjxl*.dll"`
+                jxl_lib=`find_lib "libjxl(-[0-9]+)?.dll"`
```
sezero, line 14944: I don't know regex much but your regex seems correct. However the issue is with actually `find_lib` therefore it doesn't find anything with this.

madebr: I looked into `find_lib` and I think it is supposed to be a glob, no regex.

configure.ac
```diff
@@ -438,7 +438,7 @@ if test x$enable_jxl = xyes; then
                 jxl_lib=[`find_lib libjxl.dylib`]
                 ;;
             *-*-cygwin* | *-*-mingw*)
-                jxl_lib=[`find_lib "libjxl*.dll"`]
+                jxl_lib=[`find_lib "libjxl.dll"`]
```
madebr, line 441: I settled for the simple `libjxl.dll` since `libjxl(-[0-9]+)?.dll` does not work.

.github/workflows/main.yml
```diff
@@ -190,30 +193,32 @@ jobs:
...
+        mkdir build
         ./autogen.sh
-        # We do the CMake build with stb_image and the Autotools build without
-        set -- --disable-stb-image --prefix=$(pwd)/prefix_autotools
-        # mingw-w64-*-SDL2 doesn't have SDL_test, so only build this on Unix
-        if [ "${{ matrix.platform.shell }}" = 'sh' ]; then
-          set -- "$@" --enable-tests
-          set -- "$@" --enable-installed-tests
+        if [ "x${{ runner.os }}" = "xmacOS" ]; then
+          export LDFLAGS=-Wl,-rpath,${{ steps.sdl.outputs.prefix }}/lib
```
madebr, line 203: I have to manually add the runpath of SDL2 because macOS' SIP ignores the `DYLD_LIBRARY_PATH` set by `libsdl-org/setup-sdl`.
If I don't do this, running the build-time tests fail with:
```
 FAIL: testimage
===============

dyld[7512]: Library not loaded: '@rpath/libSDL2-2.0.0.dylib'
  Referenced from: '/Users/runner/work/SDL_image/SDL_image/build/test/.libs/testimage'
  Reason: tried: '/Users/runner/work/SDL_image/SDL_image/build/.libs/libSDL2-2.0.0.dylib' (no such file), '/usr/local/lib/libSDL2-2.0.0.dylib' (no such file), '/usr/lib/libSDL2-2.0.0.dylib' (no such file)
FAIL testimage (exit status: 134)
```

CMake always uses runpath in the build directory, and removes it at install time.

external/libwebp
sezero, line 1: Looks like https://github.com/libsdl-org/SDL_image/commit/f7169f3b09f97dbfa316caaf86b7d5acad9df7ed broke this. That commit also seems to change zlib but no diff is displayed in github interface.

In any case, the external changes should go in independent of this pr, IMO.

madebr: Good eyes. I didn't notice the git submodule "revert" in f7169f3

sezero (approved):

configure.ac
sezero: What does the cmake'ry look for in mingw cases?

madebr: It first does a normal `find_library` to find the import library, which is also used for non-shared 3rd party libraries.
Then it uses [`dlltool --identify`](https://github.com/libsdl-org/SDL_image/blob/77a9125036a6f1ed272588cd8b1607e9b37552ca/cmake/PrivateSdlFunctions.cmake#L42) to identify the accompanying dll

sezero: That's good actually. Can it not be used in autotools side, I wonder.

madebr: I don't see why not.

CMake knows the path of the import library.
I think autotools will need extra logic to extract this from the output strings of  the `PKG_CHECK_MODULES ` macro.

sezero: BTW, you even do a similar thing for Mac by picking and poking the second lines below:
```sh
$ PATH=/opt/cross_osx-x86/bin:$PATH i686-apple-darwin9-otool -L lib/libxmp.dylib|grep libxmp
lib/libxmp.dylib:
	/usr/local/lib/libxmp.4.dylib (compatibility version 4.0.0, current version 4.6.0)
$ PATH=/opt/cross_osx-x86/bin:$PATH i686-apple-darwin9-otool -L lib/libxmp.4.dylib|grep libxmp
lib/libxmp.4.dylib:
	/usr/local/lib/libxmp.4.dylib (compatibility version 4.0.0, current version 4.6.0)
```

sezero:
Are there any parts from here that we'd want in SDL3 side?

sezero:
And also SDL2_mixer and SDL3_mixer..

madebr:
> Are there any parts from here that we'd want in SDL3 side?

No, I already changed ci of SDL3_image to make use of libsdl-org/setup-sdl.

> And also SDL2_mixer and SDL3_mixer..

For SDL2_mixer, when its ci switches to SDL from libsdl-org/setup-sdl.

PR #375, opened 2023-07-14
test: Copy compile definitions from the shared library

smcv:
Otherwise we won't have the definitions like -DLOAD_BMP that tell us whether each format is meant to be supported, which this test needs to know.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -65,6 +65,7 @@ set(TESTS_ENVIRONMENT
 
 foreach(prog IN LISTS ALL_TESTS)
     target_compile_definitions(${prog} PRIVATE
+        $<TARGET_PROPERTY:${sdl3_image_target_name},COMPILE_DEFINITIONS>
         "SDL_IMAGE_SAVE_JPG=$<BOOL:${SDL3IMAGE_JPG_SAVE}>"
         "SDL_IMAGE_SAVE_PNG=$<BOOL:${SDL3IMAGE_PNG_SAVE}>"
     )


test/CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ set(TESTS_ENVIRONMENT
 
 foreach(prog IN LISTS ALL_TESTS)
     target_compile_definitions(${prog} PRIVATE
+        $<TARGET_PROPERTY:${sdl3_image_target_name},COMPILE_DEFINITIONS>
```
madebr, line 68: Do we need the tests to build against an installed SDL3_image prefix?

smcv: I don't know. Do we? My only use-case for these tests has been building SDL_image and its tests together as a single monolithic project, either for build-time tests (`make check`) or "as-installed" tests in `/usr/libexec/installed-tests/SDL*_image` (which we [actively use in Debian](https://ci.debian.net/packages/libs/libsdl2-image/testing/amd64/) to check for regressions, both in SDL_image itself and in its dependencies like libpng).

With the way these tests work, I don't think they can really work correctly if built as a separate project: they want to assert that SDL3_image supports all the image formats it was meant to support, but that can't work if we don't know which image formats it was meant to support.

smcv: For instance, one of the possible bugs that I'm using these tests to prevent is: "I thought I had enabled WebP support, but I got something wrong in the packaging and now WebP is accidentally disabled, breaking games that relied on it".

madebr: Your use case is valid.

`SDL3Config.cmake` [exports various variables](https://github.com/libsdl-org/SDL_image/blob/a20997a83e8075a5595ab105470d1df1989f8786/cmake/SDL3_imageConfig.cmake.in#L11-L36).
Can these variables be used, or are some missing?

smcv: It's a bit unfortunate that we'd have to repeat ourselves in test/CMakeLists.txt with things like

```
if(SDL3IMAGE_TGA)
    target_compile_definitions(${prog} PRIVATE LOAD_TGA)
endif()
```

but if that's what you'd prefer, then yes I think that's doable.

madebr: That's true.
I made the tests work with an installed SDL_image because you did the same thing to the main SDL project.

smcv: > I made the tests work with an installed SDL_image because you did the same thing to the main SDL project

I think I only did it that way in the main SDL project because that was historically the only thing that the Autotools build supported :-)

madebr: Let's remove the possibility to build the test projects as a standalone project for all SDL (sub) projects.
The only thing you'd lose is testing an installed static SDL library, which is ok.

smcv: > Let's remove the possibility to build the test projects as a standalone project for all SDL (sub) projects

I'd be happy with that: it's how most projects' test suites work.

The `LOAD_PNG`, etc. definitions will need to be transferred from the top-level project into the test somehow, either this PR or some alternative.

PR #382, opened 2023-08-09
Update test/template.test.in for build system changes

smcv:
The build system switched from naming the executable ${exe} to naming it ${TARGET}.

Fixes: 9ed663af "cmake: use add_sdl_image_test_executable function"


madebr:
Thanks. I forgot about those files.

Issue #383, opened 2023-09-05
upgrade vendored libavif to 1.0.1 release (and dav1d to 1.2.1 release)


sezero:
For both SDL2 and SDL3.

For libavif, the new source going from 0.10.1 to 1.0.1 seems to be exif.c

For dav1d, things may be more complex.

kmilos:
Btw, this requires changes to `find_package(libavif ...)` (for both SDL2 and main) - because of `SameMajorVersion` libavif CMake config file policy 1.0.x will currently get rejected if 0.x.y was requested.

sezero:
@kmilos: libavif cmake'ry outputs `libavif.dll` w/o any versioning in dll name (like `libavif-16.dll`), is that what you mean?

kmilos:
No, I mean e.g. `find_package(libavif 0.9.2)` will FAIL if you have libavif 1.0.1 installed [because of `SameMajorVersion` policy provided by the libavif 1.0.1 CMake config itself](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file), on any platform.

So you either have to `find_package(libavif)` w/o version and then manually check `libavif_VERSION` if you want, or add an additional `find_package(libavif 1.0.0)` call if `find_package(libavif 0.9.2)` failed.

This has nothing to do w/ DLL names.

sezero:
> No, I mean e.g. `find_package(libavif 0.9.2)` will FAIL if you have libavif 1.0.1 installed [because of `SameMajorVersion` policy provided by the libavif 1.0.1 CMake config itself](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file), on any platform.
> 
> ```
> CMake Warning at CMakeLists.txt:5 (find_package):
>   Could not find a configuration file for package "libavif" that is
>   compatible with requested version "0.9.2".
> 
>   The following configuration files were considered but not accepted:
> 
>     C:/msys64/ucrt64/lib/cmake/libavif/libavif-config.cmake, version: 1.0.1
> ```
> 
> So you either have to `find_package(libavif)` w/o version and then manually check `libavif_VERSION` if you want, or add an additional `find_package(libavif 1.0.0)` call if `find_package(libavif 0.9.2)` failed.
> 

Ah, you are right. I had tested on the SDL2 branch using autotools, but not with cmake:
it fails exactly as you described.

[...]

PR #384, opened 2023-09-10
Bump vendored libraries

madebr:
- libwebp: 1.0.3 -> 1.3.2
- zlib: 1.2.12 -> 1.3
- libavif: 0.10.1 -> 1.0.1
- dav1d: 1.0.0 -> 1.2.1

And add support for building a vendored libavif library with cmake.


sezero:
Haven't read any details, but

>  libwebp: 1.0.3 -> 1.3.1

I hope libsharpyuv in new libwebp won't be a problem, i.e. libwebp and libwebpdemux libs hopefully not rely on it?

CMakeLists.txt
```diff
@@ -354,20 +367,40 @@ if(SDL3IMAGE_ZLIB)
     endif()
 endif()
 
+if(SDL3IMAGE_DAV1D)
```
sezero, line 370: Should this not be cover under the `if(SDL3IMAGE_AVIF_VENDORED)` block?

sezero:
>  dav1d: 1.0.0 -> 1.2.1

Looked very briefly at the new v1.2.1-SDL branch: You added cmake support to it, but I couldn't see how you handle nasm requirement?

sezero:
>  libwebp: 1.0.3 -> 1.3.1
> 
>  zlib: 1.2.12 -> 1.3
> 
>  libavif: 0.10.1 -> 1.0.1
> 
>  dav1d: 1.0.0 -> 1.2.1

The `.gitmodules` file should reflect the new branches

sezero:
This PR is also a good opportunity for us to audit symbol visibility from 3rd party libs. I.e.: If we statically link to them, their symbols should not be exported at all by our library.

sezero:
Great branch name btw :)

madebr:
> Haven't read any details, but
> 
> > libwebp: 1.0.3 -> 1.3.1
> 
> I hope libsharpyuv in new libwebp won't be a problem, i.e. libwebp and libwebpdemux libs hopefully not rely on it?

Looks like `libwebp.dll` depends on it, but `libwebpdemux.dll` don't:
```
>dumpbin /imports libwebp.dll | findstr dll

Dump of file libwebp.dll
    libsharpyuv.dll
    KERNEL32.dll
    VCRUNTIME140D.dll
    ucrtbased.dll
>dumpbin /imports libwebpdemux.dll | findstr dll
Dump of file libwebpdemux.dll
    libwebp.dll
    VCRUNTIME140D.dll
    ucrtbased.dll
    KERNEL32.dll
```

I don't see a configure option to disable it in the cmake script, and neither in the `configure.ac` script.

> EDIT: .. and the webp bump is not that necessary IMO, but that's just me.

Let's treat this as a yearly update.

> The `.gitmodules` file should reflect the new branches

Fixed. Thanks!


> > dav1d: 1.0.0 -> 1.2.1
> 
> Looked very briefly at the new v1.2.1-SDL branch: You added cmake support to it, but I couldn't see how you handle nasm requirement?

There is no asm support. `ENABLE_ASM` is always disabled. This was also done by the android project.
I think this is the only core feature/test I didn't port from the meson project.

> EDIT: Is Android.mk in new branch tested

No. With the `libsdl-org/setup-sdl` action, we can also test non-standard configurations.

CMakeLists.txt
madebr: You're right it would be equivalent to put it under the `if(SDL3IMAGE_AVIF_VENDORED)`, but I didn't do it for cosmatic reasons and consistency. The script also does this for zib.

sezero:
> > Haven't read any details, but
> > > libwebp: 1.0.3 -> 1.3.1
> > 
> > I hope libsharpyuv in new libwebp won't be a problem, i.e. libwebp and libwebpdemux libs hopefully not rely on it?
> 
> Looks like `libwebp.dll` depends on it, 

That's bad, actually. If we are to build vendored webp as shared lib, we should
somehow change sharpyuv to build only as static: I don't think that we wanta a
3rd *.so dependency.

> > EDIT: .. and the webp bump is not that necessary IMO, but that's just me.
> 
> Let's treat this as a yearly update.

I'm not against it.  But here you have additional issues to solve and the
new version nothing new to offer specifically to us. 
> > > dav1d: 1.0.0 -> 1.2.1
> > 
> > Looked very briefly at the new v1.2.1-SDL branch: You added cmake support to it, but I couldn't see how you handle nasm requirement?
> 
> There is no asm support. `ENABLE_ASM` is always disabled. This was also done by the android project. I think this is the only core feature/test I didn't port from the meson project.

Well, we won't expect much from dav1d side then with asm disabled..

> > EDIT: Is Android.mk in new branch tested
> 
> No. With the `libsdl-org/setup-sdl` action, we can also test non-standard configurations.

Its asm file lists / configs most possibly need updating, but after seeing the no-asm above, well...

sezero:
> > The `.gitmodules` file should reflect the new branches
> 
> Fixed. Thanks!

You seem to have reverted it..

CMakeLists.txt
sezero: >  The script also does this for zib.

Then zlib handling should be changed too, IMO

sezero:
Ragarding dav1d and cmake portage issues and no asm: If it's a big hassle, there is the option to use [libaom](https://aomedia.googlesource.com/aom/) ?

madebr:
> Ragarding dav1d and cmake portage issues and no asm: If it's a big hassle, there is the option to use [libaom](https://aomedia.googlesource.com/aom/) ?

There are 4 options:

- AOM codec
- dav1d codec
- libgav1 codec
- AVM (AV2) codec

sezero:
> > Ragarding dav1d and cmake portage issues and no asm: If it's a big hassle, there is the option to use [libaom](https://aomedia.googlesource.com/aom/) ?
> 
> There are 4 options:
> 
>     * AOM codec
>     * dav1d codec
>     * libgav1 codec
>     * AVM (AV2) codec

* AV2 is of no interest, at least for now.

* libgav1 is c++ only (ew) - let's not.

* dav1d is fast with its x86 / x86_64 / arm / neon asm.

* aom has asm optimizations too for those cpus but in compiler intrinsics as far as I remember. (its x86 32 bit detection is broken in cmake - you have to tell it to it manually)

madebr:
I added assembly support to dav1d, but it's not going to work for arm and msvc: it needs gas-preprocess.pl.

https://github.com/libsdl-org/dav1d/blob/769bd1457a9567da5e50b8c670e49234b5bea8c0/meson.build#L445-L466

I'm not 100% sure, but I think it's this: https://github.com/libav/gas-preprocessor

sezero:
Re: libwebp: If we are to bump to 1.3.1, please rebase 1.3.1-SDL branch to current 1.3.1, or cherry-pick https://github.com/libsdl-org/libwebp/commit/2af26267cdfcb63a88e5c74a85927a12d6ca1d76 into it

madebr:
I can now build dav1d with the assembly code (targeting arm64 on Android). I think `gas-preprocess.pl` is for older assemblers.
I would ignore this. (Eventually add a message to the dav1d cmake script warning people of this.)

> Re: libwebp: If we are to bump to 1.3.1, please rebase 1.3.1-SDL branch to current 1.3.1, or cherry-pick [libsdl-org/libwebp@2af2626](https://github.com/libsdl-org/libwebp/commit/2af26267cdfcb63a88e5c74a85927a12d6ca1d76) into it

I cherry-picked that commit.
I think I am already using 1.3.1-SDL. Or am I doing something wrong in this pr?

sezero:
> I can now build dav1d with the assembly code (targeting arm64 on Android). I think `gas-preprocess.pl` is for older assemblers. I would ignore this. (Eventually add a message to the dav1d cmake script warning people of this.)

OK
> > Re: libwebp: If we are to bump to 1.3.1, please rebase 1.3.1-SDL branch to current 1.3.1, or cherry-pick [libsdl-org/libwebp@2af2626](https://github.com/libsdl-org/libwebp/commit/2af26267cdfcb63a88e5c74a85927a12d6ca1d76) into it
> 
> I cherry-picked that commit.
> I think I am already using 1.3.1-SDL. Or am I doing something wrong in this pr?

Nothing  wrong

sezero:
Latest libwebp is now 1.3.2 (functionally equivalent to our 1.3.1 branch which has the necessary patch cherry-picked.)

DAV1D: Issue with nasm detection:
- You use `--version` which requires newer nasm. `-v` can be used.
  (you possibly wanted to allow yasm as a replacement?)
- The detection fails if the nasm version in use is a `rc` version, e.g. `2.16.02rc1`:
```sh
$ nasm -v
NASM version 2.16.02rc1 compiled on Sep 20 2023
$ nasm --version
NASM version 2.16.02rc1 compiled on Sep 20 2023
```

madebr:
I bumped libwebp to 1.3.2, and don't require a space after the version string anymore.

sezero:
> I bumped libwebp to 1.3.2,

OK,

> and don't require a space after the nasm version string anymore.

Works now, thanks

> I only tested with nasm. Since yasm aims to be nasm compatible, it should also work.

Well, yasm responds to `--version`, nasm responds to `-v` (and to `--version` in newer versions)
The meson build system in dav1d only tests with `-v` so I guess they won't accept yasm

One thing with dav1d: Only builds static library. Intentional?

slouken:
> One thing with dav1d: Only builds static library. Intentional?

Yes, libavif includes it statically.

madebr:
> One thing with dav1d: Only builds static library. Intentional?

For my ease, yes. And I was also "scared" of messing up the versioning.
But looks like adding shared support is easy (and generates similar libraries as those distributed by my distro).

So dav1d is now also avaliable as a shared library.

sezero:
> I bumped libwebp to 1.3.2

Branch name in `.gitmodules` need updating too. (BTW, I just deleted the obsolete 1.3.1-SDL)

sezero:
> Should I revert my dav1d change?

Why ???

madebr:
> > Should I revert my dav1d change?
> 
> Why ???

Because of Sam's comment.
```
$ ldd external/libavif/libavif.so.16
        linux-vdso.so.1 (0x00007fff75d6d000)
        libdav1d.so.6 => /home/maarten/programming/SDL_image/cmake-build-debug-vendored/external/dav1d/libdav1d.so.6 (0x00007fc237000000)
        libsharpyuv.so.0 => /lib64/libsharpyuv.so.0 (0x00007fc237360000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fc23727f000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fc236e22000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fc2373bd000)
```
I still need to remove the dependency of `libavif.so.16` on `libsharpyuv.so.o`.

slouken:
> > > Should I revert my dav1d change?
> > 
> > 
> > Why ???
> 
> Because of Sam's comment.

No, it's not a requirement, just the way the default configuration is set up.

sezero:
> ```
> $ ldd external/libavif/libavif.so.16
>         linux-vdso.so.1 (0x00007fff75d6d000)
>         libdav1d.so.6 => /home/maarten/projects/SDL_image/cmake-build-debug-vendored/external/dav1d/libdav1d.so.6 (0x00007fc237000000)
>         libsharpyuv.so.0 => /lib64/libsharpyuv.so.0 (0x00007fc237360000)
>         libm.so.6 => /lib64/libm.so.6 (0x00007fc23727f000)
>         libc.so.6 => /lib64/libc.so.6 (0x00007fc236e22000)
>         /lib64/ld-linux-x86-64.so.2 (0x00007fc2373bd000)
> ```
> 
> I still need to remove the dependency of `libavif.so.16` on `libsharpyuv.so.o`.

If you can, (i) try changing libwebp to build sharpyuv static-only, and (ii) link sharpyuv to libavif statically

BTW, are we building vendored libwebp statically? If we do, it is possible that our SDL_image dll or so have webp symbols exported: We should define `WEBP_EXTERN` and `SHARPYUV_EXTERN` as empty macros in static build cases

sezero:
More webp: Its `WEBP_EXTERN` (and `SHARPYUV_EXTERN`) actually seem broken because it doesn't take mingw into account? I'd suggest something like the following but cmake side need a corresponding surgery to define `WEBP_DLL` always for win32 but privately only to shared builds. (Not to mention its autotools...)
```diff
diff --git a/src/webp/types.h b/src/webp/types.h
index f255432..99a2209 100644
--- a/src/webp/types.h
+++ b/src/webp/types.h
@@ -41,10 +41,12 @@ typedef long long int int64_t;
 // signature for e.g., Windows DLL builds.
-# if defined(__GNUC__) && __GNUC__ >= 4
+#  if defined(_WIN32)
+#   if defined(WEBP_DLL)
+#    define WEBP_EXTERN __declspec(dllexport)
+#   else
+#    define WEBP_EXTERN extern
+#  endif
+# elif defined(__GNUC__) && __GNUC__ >= 4
 #  define WEBP_EXTERN extern __attribute__ ((visibility ("default")))
 # else
-#  if defined(_MSC_VER) && defined(WEBP_DLL)
-#   define WEBP_EXTERN __declspec(dllexport)
-#  else
 #   define WEBP_EXTERN extern
-#  endif
 # endif  /* __GNUC__ >= 4 */
```

madebr:
I don't think mingw needs `__declspec(dllexport)` (on my system at least).
Consider the following source:
```c
int __attribute__((visibility("default"))) some_function(void);
int  some_function(void) { return 1337; }
```
compiled and linked with `x86_64-w64-mingw32-gcc -O3 -s src.c -shared -o somelib.dll -Wl,--out-implib,libsomelib.a`.


Then, `x86_64-w64-mingw32-objdump -x somelib.dll` contains:
```
[Ordinal/Name Pointer] Table
        [   0] some_function
```
and `x86_64-w64-mingw32-nm libsomelib.a` contains:
```
0000000000000000 T some_function
```

madebr:
> BTW, are we building vendored libwebp statically? If we do, it is possible that our SDL_image dll or so have webp symbols exported: We should define `WEBP_EXTERN` and `SHARPYUV_EXTERN` as empty macros in static build cases

No, everything is build as a shared library.

The SDL3_image install prefix, when building the vendored libraries as shared libraries creates the following install prefix:
```
prefix
├── include
│   ├── decode.h
│   ├── demux.h
│   ├── encode.h
│   ├── mux_types.h
│   ├── SDL3_image
│   │   └── SDL_image.h
│   └── types.h
├── lib64
│   ├── cmake
│   │   └── SDL3_image
│   │       ├── Findlibjxl.cmake
│   │       ├── Findwebp.cmake
│   │       ├── SDL3_imageConfig.cmake
│   │       ├── SDL3_imageConfigVersion.cmake
│   │       ├── SDL3_image-shared-targets.cmake
│   │       └── SDL3_image-shared-targets-debug.cmake
│   ├── libavif.so -> libavif.so.16
│   ├── libavif.so.16 -> libavif.so.16.0.1
│   ├── libavif.so.16.0.1
│   ├── libbrotlicommon.so -> libbrotlicommon.so.1
│   ├── libbrotlicommon.so.1 -> libbrotlicommon.so.1.0.9
│   ├── libbrotlicommon.so.1.0.9
│   ├── libbrotlidec.so -> libbrotlidec.so.1
│   ├── libbrotlidec.so.1 -> libbrotlidec.so.1.0.9
│   ├── libbrotlidec.so.1.0.9
│   ├── libbrotlienc.so -> libbrotlienc.so.1
│   ├── libbrotlienc.so.1 -> libbrotlienc.so.1.0.9
│   ├── libbrotlienc.so.1.0.9
│   ├── libdav1d.so -> libdav1d.so.6
│   ├── libdav1d.so.6 -> libdav1d.so.6.9.0
│   ├── libdav1d.so.6.9.0
│   ├── libjxl.so -> libjxl.so.0.6
│   ├── libjxl.so.0.6 -> libjxl.so.0.6.1
│   ├── libjxl.so.0.6.1
│   ├── libSDL3_image.so -> libSDL3_image.so.0
│   ├── libSDL3_image.so.0 -> libSDL3_image.so.0.0.0
│   ├── libSDL3_image.so.0.0.0
│   ├── libtiff.so -> libtiff.so.6
│   ├── libtiff.so.6 -> libtiff.so.6.0.1
│   ├── libtiff.so.6.0.1
│   ├── libwebpdemux.so -> libwebpdemux.so.2
│   ├── libwebpdemux.so.2 -> libwebpdemux.so.2.0.14
│   ├── libwebpdemux.so.2.0.14
│   ├── libwebp.so -> libwebp.so.7
│   ├── libwebp.so.7 -> libwebp.so.7.1.8
│   ├── libwebp.so.7.1.8
│   └── pkgconfig
│       └── sdl3-image.pc
├── libexec
│   └── installed-tests
│       └── SDL3_image
│           ├── palette.bmp
│           ├── palette.gif
│           ├── sample.avif
│           ├── sample.bmp
│           ├── sample.cur
│           ├── sample.ico
│           ├── sample.jpg
│           ├── sample.jxl
│           ├── sample.pcx
│           ├── sample.png
│           ├── sample.pnm
│           ├── sample.qoi
│           ├── sample.tga
│           ├── sample.tif
│           ├── sample.webp
│           ├── sample.xcf
│           ├── sample.xpm
│           ├── svg64.bmp
│           ├── svg.bmp
│           ├── svg-class.bmp
│           ├── svg-class.svg
│           ├── svg.svg
│           └── testimage
└── share
    ├── installed-tests
    │   └── SDL3_image
    │       └── testimage.test
    ├── licenses
    │   └── SDL3_image
    │       └── LICENSE.txt
    └── man
        └── man3
            ├── IMG_FreeAnimation.3
            ├── IMG_Init.3
            ├── IMG_isAVIF.3
            ├── IMG_isBMP.3
            ├── IMG_isCUR.3
            ├── IMG_isGIF.3
            ├── IMG_isICO.3
            ├── IMG_isJPG.3
            ├── IMG_isJXL.3
            ├── IMG_isLBM.3
            ├── IMG_isPCX.3
            ├── IMG_isPNG.3
            ├── IMG_isPNM.3
            ├── IMG_isQOI.3
            ├── IMG_isSVG.3
            ├── IMG_isTIF.3
            ├── IMG_isWEBP.3
            ├── IMG_isXCF.3
            ├── IMG_isXPM.3
            ├── IMG_isXV.3
            ├── IMG_Linked_Version.3
            ├── IMG_Load.3
            ├── IMG_LoadAnimation.3
            ├── IMG_LoadAnimation_RW.3
            ├── IMG_LoadAnimationTyped_RW.3
            ├── IMG_LoadAVIF_RW.3
            ├── IMG_LoadBMP_RW.3
            ├── IMG_LoadCUR_RW.3
            ├── IMG_LoadGIFAnimation_RW.3
            ├── IMG_LoadGIF_RW.3
            ├── IMG_LoadICO_RW.3
            ├── IMG_LoadJPG_RW.3
            ├── IMG_LoadJXL_RW.3
            ├── IMG_LoadLBM_RW.3
            ├── IMG_LoadPCX_RW.3
            ├── IMG_LoadPNG_RW.3
            ├── IMG_LoadPNM_RW.3
            ├── IMG_LoadQOI_RW.3
            ├── IMG_Load_RW.3
            ├── IMG_LoadSizedSVG_RW.3
            ├── IMG_LoadSVG_RW.3
            ├── IMG_LoadTexture.3
            ├── IMG_LoadTexture_RW.3
            ├── IMG_LoadTextureTyped_RW.3
            ├── IMG_LoadTGA_RW.3
            ├── IMG_LoadTIF_RW.3
            ├── IMG_LoadTyped_RW.3
            ├── IMG_LoadWEBPAnimation_RW.3
            ├── IMG_LoadWEBP_RW.3
            ├── IMG_LoadXCF_RW.3
            ├── IMG_LoadXPM_RW.3
            ├── IMG_LoadXV_RW.3
            ├── IMG_Quit.3
            ├── IMG_ReadXPMFromArray.3
            ├── IMG_ReadXPMFromArrayToRGB888.3
            ├── IMG_SaveJPG.3
            ├── IMG_SaveJPG_RW.3
            ├── IMG_SavePNG.3
            └── IMG_SavePNG_RW.3

17 directories, 127 files
```

The extra include headers are a consequence of dependencies doing `PUBLIC_HEADER`, and can be patched out.
But I left it as a don't care, and also to diminish the number of patches.

sezero:
Missing case in your test -- consider the following:
```c
int __attribute__((visibility("default"))) some_function(void);
int  some_function(void) { return 1337; }
int  more_function(void) { return 0; }
```

Built like this:
```sh
x86_64-w64-mingw32-gcc -O3 -Wall -fvisibility=hidden -s src.c -shared -o somelib.dll -Wl,--out-implib,libsomelib.a
src.c: In function 'more_function':
src.c:3:1: warning: visibility attribute not supported in this configuration; ignored
```

... the build happily exports more_function for me.

madebr:
> If you can, (i) try changing libwebp to build sharpyuv static-only, and (ii) link sharpyuv to libavif statically

I applied that change.
The dependency needs no extra check, because both libraries are either both vendored, or both system libraries.

madebr:
> ```shell
> x86_64-w64-mingw32-gcc -O3 -Wall -fvisibility=hidden -s src.c -shared -o somelib.dll -Wl,--out-implib,libsomelib.a
> src.c: In function 'more_function':
> src.c:3:1: warning: visibility attribute not supported in this configuration; ignored
> ```

Mine doesn't warn. (x86_64-w64-mingw32-gcc (GCC) 12.2.1 20221121)

sezero:
> > ```shell
> > x86_64-w64-mingw32-gcc -O3 -Wall -fvisibility=hidden -s src.c -shared -o somelib.dll -Wl,--out-implib,libsomelib.a
> > src.c: In function 'more_function':
> > src.c:3:1: warning: visibility attribute not supported in this configuration; ignored
> > ```
> 
> Mine doesn't warn, and doesn't export the `more_function` symbol. (x86_64-w64-mingw32-gcc (GCC) 12.2.1 20221121) But feel free to apply this change (after my pr please :P)

I don't intend to blindly apply the patch for the moment, so no worries :)

slouken:
> The extra include headers are a consequence of dependencies doing `PUBLIC_HEADER`, and can be patched out. But I left it as a don't care, and also to diminish the number of patches.

We should not install public headers from vendored libraries. Aside from user confusion, this would cause our package to potentially conflict with already installed packages.

madebr:
> We should not install public headers from vendored libraries. Aside from user confusion, this would cause our package to potentially conflict with already installed packages.

Fixed. This turned out to be easier then I thought. No patch required.

sezero:
I don't know what else is left, but one thing about libavif: We want to
build it with `CMAKE_DLL_NAME_WITH_SOVERSION=ON` -- see:
https://github.com/libsdl-org/SDL_image/issues/383#issuecomment-1711789405

madebr:
> I don't know what else is left, but one thing about libavif: We want to build it with `CMAKE_DLL_NAME_WITH_SOVERSION=ON` -- see: [#383 (comment)](https://github.com/libsdl-org/SDL_image/issues/383#issuecomment-1711789405)

[That property is CMake 3.27+ only](https://cmake.org/cmake/help/latest/prop_tgt/DLL_NAME_WITH_SOVERSION.html).
I rolled my own versioning with:
```cmake
if(SDL3IMAGE_AVIF_SHARED AND WIN32 AND NOT CMAKE_PLATFORM_NO_VERSIONED_SONAME)
    set_property(TARGET avif PROPERTY OUTPUT_NAME "avif-$<TARGET_PROPERTY:avif,SOVERSION>")
    set_property(TARGET avif PROPERTY DLL_NAME_WITH_SOVERSION OFF)
endif()
```

madebr:
This builds for me on Linux and mingw, and the tests pass (using Wine on Linux).

sezero:
> > I don't know what else is left, but one thing about libavif: We want to build it with `CMAKE_DLL_NAME_WITH_SOVERSION=ON` -- see: [#383 (comment)](https://github.com/libsdl-org/SDL_image/issues/383#issuecomment-1711789405)
> 
> [That property is CMake 3.27+ only](https://cmake.org/cmake/help/latest/prop_tgt/DLL_NAME_WITH_SOVERSION.html). I rolled my own versioning with:
> 
> ```cmake
> if(SDL3IMAGE_AVIF_SHARED AND WIN32 AND NOT CMAKE_PLATFORM_NO_VERSIONED_SONAME)
>     set_property(TARGET avif PROPERTY OUTPUT_NAME "avif-$<TARGET_PROPERTY:avif,SOVERSION>")
>     set_property(TARGET avif PROPERTY DLL_NAME_WITH_SOVERSION OFF)
> endif()
> ```

Well, if it works and outputs libavif-16.dll instead of libavif.dll then it's fine.
(Although I wonder whether we should/could've changed libavif for it..)

madebr:
> Well, if it works and outputs libavif-16.dll instead of libavif.dll then it's fine. (Although I wonder whether we should/could've changed libavif for it..)

You're right. I patched libavif's cmake script.

sezero:
Anything else left?

Maybe a few things noted at #384 but they are mostly SDL2-related and can
be put in a new pull request. 

(I just rebuilt the libavif dlls for the msvc project and uploaded, btw)

slouken:
> > Well, if it works and outputs libavif-16.dll instead of libavif.dll then it's fine. (Although I wonder whether we should/could've changed libavif for it..)
> 
> You're right. I patched libavif's cmake script.

We should link with whatever the default DLL name is for each project. That way if someone builds it independently, they can drop in a new DLL.

sezero:
> > > Well, if it works and outputs libavif-16.dll instead of libavif.dll then it's fine. (Although I wonder whether we should/could've changed libavif for it..)
> > 
> > 
> > You're right. I patched libavif's cmake script.
> 
> We should link with whatever the default DLL name is for each project. That way if someone builds it independently, they can drop in a new DLL.

I disagree: We shouldn't. At the very least, the default windows builds of libavif have no versioning in their names, only libavif.dll, and 0.9.3, 0.10, 0.11 and 1.0 are abi-incompatible.

slouken:
> > > > Well, if it works and outputs libavif-16.dll instead of libavif.dll then it's fine. (Although I wonder whether we should/could've changed libavif for it..)
> > > 
> > > 
> > > You're right. I patched libavif's cmake script.
> > 
> > 
> > We should link with whatever the default DLL name is for each project. That way if someone builds it independently, they can drop in a new DLL.
> 
> I disagree: We shouldn't. At the very least, the default windows builds of libavif have no versioning in their names, only libavif.dll, and 0.9.3, 0.10, 0.11 and 1.0 are abi-incompatible.

Fair enough.

madebr:
msys2 is not consistent in versioning their libraries.
[dav1d](https://packages.msys2.org/package/mingw-w64-x86_64-dav1d)'s dll contains no versioning.

sezero:
zlib tagged new 1.3.1 release: Do we want to upgrade?

slouken:
Sure. In general we want to follow library public releases.

sezero:
OK, I created v1.3.1-SDL branch in our zlib fork and applied two of our patches.
Intentionally not applied https://github.com/libsdl-org/zlib/commit/8d8abbc276d1151ca9df98ae6ee4638a7fe6df25:
zlib 1.3.1 has a new `ZLIB_BUILD_EXAMPLES` option, which we can disable in
our own cmake'ry: https://github.com/libsdl-org/zlib/blob/master/CMakeLists.txt#L8

@madebr: Will you do the upgrade in SDL_image or should I attempt it?

@slouken: Should we upgrade in 2.8.x release branch too?

madebr:
Feel free to do it, you probably need to add `set(ZLIB_BUILD_EXAMPLES OFF CACHE BOOL "zlib examples")`.

slouken:
What are the changes in 1.3.1? If it's security fixes, we should definitely add it to the release branch. If not, it's probably fine to leave it in SDL2 for now.

sezero:
> What are the changes in 1.3.1? If it's security fixes, we should definitely add it to the release branch. If not, it's probably fine to leave it in SDL2 for now.

Changelog doesn't look critical to me: https://github.com/madler/zlib/releases/tag/v1.3.1
Leave release branch as is?

slouken:
Yep, sounds good.

sezero:
> DLL_NAME_WITH_SOVERSION

@madebr: Can we not emulate it with something like the following regardless of cmake version?

```cmake
    cmake_push_check_state()
    if(WIN32)
        set (CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION 1)
    endif()
    add_library (avif ...
    stuff...
    cmake_pop_check_state()
```

sezero:
> Older CMake versions will just ignore `CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION`.

Are you sure? It works for me with cmake 3.1.0 ??

madebr:
> > Older CMake versions will just ignore `CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION`.
> 
> Are you sure? It works for me with cmake 3.1.0 (at least with mingw) ??

No, I'm not.
I mixed `DLL_NAME_WITH_SOVERSION` with `CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION`.
But the lack of documentation of this variable makes me wary.
Its only usage is in `Modules/Platform/CYGWIN.cmake`.

sezero:
Looked at the sources:

In cmGeneratorTarget.cxx : GetFullNameInternalComponents() of 3.26.6
there is this:
```c++
  // Name shared libraries with their version number on some platforms.
  if (cmValue soversion = this->GetProperty("SOVERSION")) {
    if (this->GetType() == cmStateEnums::SHARED_LIBRARY &&
        !isImportedLibraryArtifact &&
        this->Makefile->IsOn("CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION")) {
      outBase += "-";
      outBase += *soversion;
    }
  }
```

In 3.1.0, it is like this in cmTarget.cxx : GetFullNameInternal() :
```c++
  // Name shared libraries with their version number on some platforms.
  if(const char* soversion = this->GetProperty("SOVERSION"))
    {
    if(this->GetType() == cmTarget::SHARED_LIBRARY && !implib &&
       this->Makefile->IsOn("CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION"))
      {
      outBase += "-";
      outBase += soversion;
      }
    }
```

madebr:
Just tried it out in a test project, looks like it works.
Feel free to apply this in `libsdl-org/libavif` and remove my workaround in SDL_image's cmake script.

sezero:
So, the libavif patch should look like the following ?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index dbdcd01..4b7c791 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -533,6 +533,11 @@ if(NOT AVIF_CODEC_AOM
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
+include(CMakePushCheckState)
+cmake_push_check_state()
+if(BUILD_SHARED_LIBS AND WIN32)
+    set(CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION 1)
+endif()
 add_library(avif ${AVIF_SRCS})
 set_target_properties(avif PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION} C_VISIBILITY_PRESET hidden)
 target_compile_definitions(avif PRIVATE ${AVIF_PLATFORM_DEFINITIONS} ${AVIF_CODEC_DEFINITIONS})
@@ -541,10 +546,7 @@ target_include_directories(
     avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include> PRIVATE ${AVIF_PLATFORM_INCLUDES}
                                                                                                       ${AVIF_CODEC_INCLUDES}
 )
-if(BUILD_SHARED_LIBS AND WIN32 AND NOT CMAKE_PLATFORM_NO_VERSIONED_SONAME)
-    set_property(TARGET avif PROPERTY OUTPUT_NAME "avif-${LIBRARY_SOVERSION}")
-    set_property(TARGET avif PROPERTY DLL_NAME_WITH_SOVERSION OFF)
-endif()
+cmake_pop_check_state()
 target_include_directories(avif PRIVATE "${libavif_SOURCE_DIR}/third_party/libyuv/include/")
 set(AVIF_PKG_CONFIG_EXTRA_CFLAGS "")
 if(BUILD_SHARED_LIBS)
```

Am I supposed to set `DLL_NAME_WITH_SOVERSION` to 0 to be safe??

NOTE: Built libavif patched as above using Cmake+MSVC, it gave me an `avif-16.dll` not `libavif-16.dll`, so should we set `CMAKE_SHARED_LIBRARY_PREFIX` to `lib` or something? (Your original workaround doesn't set prefix either...) Or do like that as is?

Not sure what to change in SDL_image's cmake'ry, though?? The only commit about `DLL_NAME_WITH_SOVERSION` is 69ba38c2dd2fcba246582bbc3fcbe9de48654f98 which seems irrelevant, and I can't see what to change in the avif block.

madebr:
You need to do `CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION` in the cmake script of libavif.
It looks like it is a variable that only takes affect at the end of the current cmake script (like `CMAKE_C_FLAGS` etc).
`cmake_push_check_state/cmake_pop_check_state` does nothing in this regard.

Don't disable the `DLL_NAME_WITH_SOVERSION` property, or just unconditionally enable it.
In a dummy project, `CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION` and `DLL_NAME_WITH_SOVERSION` don't stack.
Disabling `DLL_NAME_WITH_SOVERSION` will disable the SOVERSION prefix.

sezero:
OK, got it. So the final libavif patch is the following yes? (I'm keeping MSVC case as is)

And there is nothing to do in SDL_image cmake'ry?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index dbdcd01..28a67a5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -531,22 +531,21 @@ if(NOT AVIF_CODEC_AOM
    AND NOT AVIF_CODEC_AVM
 )
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
+if(BUILD_SHARED_LIBS AND WIN32)
+    set(CMAKE_SHARED_LIBRARY_NAME_WITH_VERSION 1)
+endif()
 add_library(avif ${AVIF_SRCS})
 set_target_properties(avif PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION} C_VISIBILITY_PRESET hidden)
 target_compile_definitions(avif PRIVATE ${AVIF_PLATFORM_DEFINITIONS} ${AVIF_CODEC_DEFINITIONS})
 target_link_libraries(avif PRIVATE ${AVIF_CODEC_LIBRARIES} ${AVIF_PLATFORM_LIBRARIES})
 target_include_directories(
     avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include> PRIVATE ${AVIF_PLATFORM_INCLUDES}
                                                                                                       ${AVIF_CODEC_INCLUDES}
 )
-if(BUILD_SHARED_LIBS AND WIN32 AND NOT CMAKE_PLATFORM_NO_VERSIONED_SONAME)
-    set_property(TARGET avif PROPERTY OUTPUT_NAME "avif-${LIBRARY_SOVERSION}")
-    set_property(TARGET avif PROPERTY DLL_NAME_WITH_SOVERSION OFF)
-endif()
 target_include_directories(avif PRIVATE "${libavif_SOURCE_DIR}/third_party/libyuv/include/")
 set(AVIF_PKG_CONFIG_EXTRA_CFLAGS "")
 if(BUILD_SHARED_LIBS)
     target_compile_definitions(avif PUBLIC AVIF_DLL PRIVATE AVIF_BUILDING_SHARED_LIBS)
     set(AVIF_PKG_CONFIG_EXTRA_CFLAGS " -DAVIF_DLL")
```

madebr:
> OK, got it. So the final libavif patch is the following yes? (I'm keeping MSVC case as is)

Looks fine to me (if CI generates the same set of files).
What MSVC case are you referring to?

> 
> And there is nothing to do in SDL_image cmake'ry?

I don't think so: `$<TARGET_SONAME_FILE:avif>` is used there, which should just pick up the final name.

sezero:
> > OK, got it. So the final libavif patch is the following yes? (I'm keeping MSVC case as is)
> 
> Looks fine to me (if CI generates the same set of files).

OK, will be testing patches in my fork soon

> What MSVC case are you referring to?

That MSVC is outputting `avif-16.dll`, not `libavif-16.dll` as I mentioned above

> > And there is nothing to do in SDL_image cmake'ry?
> 
> I don't think so: `$<TARGET_SONAME_FILE:avif>` is used there, which should just pick up the final name.

OK, got it, Not touching SDL_image.

madebr:
> > What MSVC case are you referring to?
> 
> That MSVC is outputting `avif-16.dll`, not `libavif-16.dll` as I mentioned above

This can be fixed by doing `set_property(TARGET avif PROPERTY PREFIX "lib")` in the avif project.

(Sorry, I didn't read that part of your original post)

sezero:
OK, pushed https://github.com/libsdl-org/libavif/commit/5bcd7d031e058b171364916a4dbef5fc5d6a28e2 and updated vendored libavif in SDL_image, both main and SDL2 branches.

libavif has a CI failure, seems unrelated to the patch.

SDL_image, both main SDL2 branches, has test failures in macos runs (for jxl I think), seems unrelated to the patch.

Downloaded the SDL3 MSVC artifact from CI runs: avif dll is named libavif-16.dll and SDL3_image.dll is looking for it with the correct name judging by the strings in the dll.

madebr:
> OK, pushed [libsdl-org/libavif@5bcd7d0](https://github.com/libsdl-org/libavif/commit/5bcd7d031e058b171364916a4dbef5fc5d6a28e2) and updated vendored libavif in SDL_image, both main and SDL2 branches.

Great!

> 
> libavif has a CI failure, seems unrelated to the patch.

I think you're seeing an issue with libjxl that's too recent.
I also encountered it in https://github.com/libsdl-org/SDL_image/pull/443#issuecomment-2035003405.
It only shows up on Macos.
My distro's libjxl is older, and our vendored libjxl is even older.

PR #386, opened 2023-09-16
Create n3ds.yml

RetroGamer02:
Adds a workflow to build for the Nintendo 3DS.


madebr:
Is it possible to integrate this into the matrix of the main workflow? That way, we don't need to maintain almost similar files.

RetroGamer02:
> Is it possible to integrate this into the matrix of the main workflow? That way, we don't need to maintain almost similar files.

Done let me know if any other changes are needed.

madebr:
I meant adding n3ds to the matrix in the main workflow, such that the steps with cmake build commands aren't duplicated.
Your latest change causes ci to no longer run.
https://github.com/libsdl-org/SDL_image/actions/runs/6204590724

madebr:
As an example, see this action I used to verify setup-sdl:

https://github.com/madebr/setup-sdl-test/blob/33322cbb4d6118f1874e8e9ca4fcb216434c5857/.github/workflows/main.yml#L68-L114

madebr:
I'm closing this issue. I'm not sure it adds extra test coverage. I'll be glad if I'm proven wrong.

PR #391, opened 2023-10-05
Use CMAKE_CURRENT_*_DIR instead of CMAKE_*_DIR

humbertodias:
This PR allows for building with cmake's [add_subdirectory](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory) function, like so:

# Build and include SDL in main project
```shell
add_subdirectory(lib/SDL2_image lib/SDL2_image_BIN EXCLUDE_FROM_ALL)
target_link_libraries(main PRIVATE SDL2_image)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ set(MINOR_VERSION 0)
 set(MICRO_VERSION 0)
 set(SDL_REQUIRED_VERSION 3.0.0)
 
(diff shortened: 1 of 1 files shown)


sezero:
(i) Same goes for the SDL2 branch

(ii) Same goes for both SDL2 nd SDL3 branches of all other SDL satellite libs

madebr:
Like the grim reaper, I went around and applied this change to all satellite libraries.
SDL itself already did this.

PR #392, opened 2023-10-14
cmake: backport support for vendored libavif to SDL2_image

sezero:
@madebr: Please review this.

There is a CI run failure with MacOS vendored build like:
```
   "_dav1d_warp_affine_8x8_8bpc_sse2", referenced from:
      _dav1d_mc_dsp_init_8bpc in mc_tmpl.c.o
```

... but I think that is a problem with your dav1d cmake support
and not with this backport: I _think_ dav1d is built with either
incorrect nasm output format flags or without any flags at all.
In mac case, we want `-fmacho`, in windows case we want `-fwin`,
and in linux/elf case we want `-felf`, and all need appending a
`32` or `64` depending on x86 or x64 target.


sezero:
OK, I cross-built dav1d for x86_64 mac, the same errors, and cmake
seems to use `-f macho64` correctly. On the other hand build using
meson as a dylib with the same environment succeeds.

The problem seems to be with symbol decoration: macho and windows-
x86 symbols must be prefixed with and underscore. This probably has
something to do with meson defining `PREFIX` macro...

sezero:
For reference, here are the configs generated by meson: [meson_configs.zip](https://github.com/libsdl-org/SDL_image/files/12905349/meson_configs.zip)

sezero:
Manually adding `PREFIX` define in config.asm and config.h generated
by your dav1d cmake port: build succeeds.

However: There seem to be more missing stuff in dav1d cmake port,
one such thing is `STACK_ALIGNMENT` in config.asm. There may be others.

Is it not possible to run meson configurator from within cmake??

sezero:
> Is it not possible to run meson configurator from within cmake??

I could find https://discourse.cmake.org/t/3985 and
https://cmake.org/cmake/help/latest/module/ExternalProject.html,
but I'm not fluent enough with cmake to make use of them

sezero:
The following seems to work for me and seems to match meson better.
Tested by cross-compiling dav1d as an x86_64 dylib. @madebr: Please
review. Notes:
- New `PREFIX` variable: Set for APPLE or WIN32+X86 if ASM is enabled
- `TRIM_DSP_FUNCTIONS` is only set for Release and MinSizeRel builds
- invented new `DAV1D_STACK_ALIGNMENT`: can be set by user on the
  cmake cmdline or decided depending on platform / cpu.
- invented new `config_variables_value`, with its value part being
  written to config.h.in as `@value_var@` -- make sure it's correct
  thing to do.
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0ee5638..2131e15 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,7 @@ else()
 endif()
 
 set(config_variables)
+set(config_variables_value)
 set(config01_variables)
 
 check_cpu_architecture(x64 "DAV1D_CPU_X64")
@@ -110,6 +111,9 @@ if(DAV1D_ASM)
     check_c_source_compiles("__asm__ (\n\".func cmake_test\"\n\".endfunc\"\n);\n" HAVE_AS_FUNC)
     list(APPEND config01_variables HAVE_AS_FUNC)
   endif()
+  if(APPLE OR (WIN32 AND DAV1D_CPU_X86))
+    set(PREFIX 1)
+  endif()
 endif()
 
 foreach(depth 8 16)
@@ -137,7 +141,20 @@ else()
 endif()
 
 list(APPEND config01_variables TRIM_DSP_FUNCTIONS)
+if(CMAKE_BUILD_TYPE STREQUAL "Release" OR
+   CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
 set(TRIM_DSP_FUNCTIONS 1)
+endif()
+
+if(NOT DAV1D_STACK_ALIGNMENT)
+  if(DAV1D_CPU_X64 OR APPLE OR (CMAKE_SYSTEM_NAME MATCHES ".*Linux"))
+    set(DAV1D_STACK_ALIGNMENT 16)
+  else()
+    set(DAV1D_STACK_ALIGNMENT 4)
+  endif()
+endif()
+list(APPEND config_variables_value STACK_ALIGNMENT)
+set(STACK_ALIGNMENT ${DAV1D_STACK_ALIGNMENT})
 
 if(WIN32)
   add_compile_definitions(
@@ -168,7 +185,7 @@ if(NOT HAVE_STDATOMIC_H)
     endif()
   endif()
 endif()
-list(APPEND config_variables HAVE_UNISTD_H HAVE_IO_H HAVE_PTHREAD_NP_H)
+list(APPEND config_variables HAVE_UNISTD_H HAVE_IO_H HAVE_PTHREAD_NP_H PREFIX)
 check_include_file(unistd.h HAVE_UNISTD_H)
 check_include_file(io.h HAVE_IO_H)
 check_include_file(pthread_np.h HAVE_PTHREAD_NP_H)
@@ -224,6 +241,9 @@ set(cfg_h)
 foreach(cfg01 IN LISTS config01_variables)
   string(APPEND cfg_h "#cmakedefine01 ${cfg01}\n")
 endforeach()
+foreach(cfg IN LISTS config_variables_value)
+  string(APPEND cfg_h "#cmakedefine ${cfg} @${cfg}@\n")
+endforeach()
 foreach(cfg IN LISTS config_variables)
   string(APPEND cfg_h "#cmakedefine ${cfg}\n")
 endforeach()
```

sezero:
OK, I pushed https://github.com/libsdl-org/dav1d/commit/66d3dd88dc79a56b1756a976f8922416ce7382ea to our dav1d 1.2.1-SDL branch and bumped the dav1d revision here accordingly. The CI link failure seems to be cured. @madebr: Please confirm that things are in order. If you confirm, I will bump it in SDL3 branch too.

@slouken: Noticed a few things in your dav1d Android configuration:

- `PREFIX` is defined in all configs: Is that accurate, or some copy/paste artifact?

- There is no config.asm in none of the configs.

- `STACK_ALIGNMENT` is defined as 32: is that deliberately picked? Because meson build systems defines it either as 4 or 16 unless user-specified.

- `STACK_ALIGNMENT` is defined only for x86_64 config but not arm configs: Intentional??

madebr:
Your dav1d changes look fine, I have these suggestions:
- Make `TRIM_DSP_FUNCTIONS` a compile definitions, instead of adding it to the config file. `CMAKE_BUILD_TYPE` is not defined for multi-config generators, such as Visual Studio.
- Make `DAV1D_STACK_ALIGNMENT` a configurable option, that can be overriden by users.


```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,19 +151,15 @@ else()
   set(CONFIG_LOG 0)
 endif()
 
-list(APPEND config01_variables TRIM_DSP_FUNCTIONS)
-if(CMAKE_BUILD_TYPE STREQUAL "Release" OR
-   CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
-set(TRIM_DSP_FUNCTIONS 1)
-endif()
+# TRIM_DSP_FUNCTIONS=1 if Release or Debug mode
+add_definitions(-DTRIM_DSP_FUNCTIONS=$<BOOL:$<OR:$<CONFIG:Release>,$<CONFIG:MinSizeRel>>>)
 
-if(NOT DAV1D_STACK_ALIGNMENT)
-  if(DAV1D_CPU_X64 OR APPLE OR (CMAKE_SYSTEM_NAME MATCHES ".*Linux"))
-    set(DAV1D_STACK_ALIGNMENT 16)
-  else()
-    set(DAV1D_STACK_ALIGNMENT 4)
-  endif()
+if(DAV1D_CPU_X64 OR APPLE OR CMAKE_SYSTEM_NAME MATCHES ".*Linux")
+    set(DAV1D_STACK_ALIGNMENT_DEFAULT 16)
+else()
+    set(DAV1D_STACK_ALIGNMENT_DEFAULT 4)
 endif()
+set(DAV1D_STACK_ALIGNMENT "${DAV1D_STACK_ALIGNMENT_DEFAULT}" CACHE STRING "Dav1d stack alignment")
 list(APPEND config_variables_value STACK_ALIGNMENT)
 set(STACK_ALIGNMENT ${DAV1D_STACK_ALIGNMENT})
```



> > Is it not possible to run meson configurator from within cmake??
> 
> I could find https://discourse.cmake.org/t/3985 and https://cmake.org/cmake/help/latest/module/ExternalProject.html, but I'm not fluent enough with cmake to make use of them

I tried this approach and asked on the #cmake channel on cpplang.slack.com, but got no satisfying answer (apart from people telling me they got no interest in cmake, or that I should use a package manager).
While trying, I had to [auto-generate a "cross build definition file" for meson](https://mesonbuild.com/Cross-compilation.html#cross-compilation) which I can't do for all toolchains + cmake gnerators.

sezero:
> Your dav1d changes look fine, I have these suggestions:
> 
>     * Make `TRIM_DSP_FUNCTIONS` a compile definitions, instead of adding it to the config file. `CMAKE_BUILD_TYPE` is not defined for multi-config generators, such as Visual Studio.

OK

>     * Make `DAV1D_STACK_ALIGNMENT` a configurable option, that can be overriden by users.
> 
> ```diff
> [...]
> +if(DAV1D_CPU_X64 OR APPLE OR CMAKE_SYSTEM_NAME MATCHES ".*Linux")
> +    set(DAV1D_STACK_ALIGNMENT_DEFAULT 16)
> +else()
> +    set(DAV1D_STACK_ALIGNMENT_DEFAULT 4)
>  endif()
> +set(DAV1D_STACK_ALIGNMENT "${DAV1D_STACK_ALIGNMENT_DEFAULT}" CACHE STRING "Dav1d stack alignment")
> ```

OK. It's guaranteed that cmdline arg like `-DDAV1D_STACK_ALIGNMENT` works here, yes?

> > > Is it not possible to run meson configurator from within cmake??
> > 
> > I could find https://discourse.cmake.org/t/3985 and https://cmake.org/cmake/help/latest/module/ExternalProject.html, but I'm not fluent enough with cmake to make use of them
> 
> I tried this approach and asked on the #cmake channel on cpplang.slack.com, but got no satisfying answer (apart from people telling me they got no interest in cmake, or that I should use a package manager). While trying, I had to [auto-generate a "cross build definition file" for meson](https://mesonbuild.com/Cross-compilation.html#cross-compilation) which I can't do for all toolchains + cmake gnerators.

OK, then.

sezero:
> OK. It's guaranteed that cmdline arg like `-DDAV1D_STACK_ALIGNMENT` works here, yes?

OK, stopped beign lazy and confirmed that it works, pushed the suggested changes to dav1d, bumped the rev here.

Is this PR good?

@slouken: Comments for the notes I wrote about your Android and Xcode configs?

cmake/PrivateSdlFunctions.cmake
```diff
@@ -258,12 +258,16 @@ function(target_get_dynamic_library DEST TARGET)
     set(${DEST} ${result} PARENT_SCOPE)
 endfunction()
 
-macro(sdl_check_project_in_subfolder relative_subfolder name vendored_option)
-    if(NOT EXISTS "${PROJECT_SOURCE_DIR}/${relative_subfolder}/CMakeLists.txt")
-        message(FATAL_ERROR "No cmake project for ${name} found in ${relative_subfolder}.\n"
+function(sdl_check_project_in_subfolder relative_subfolder name vendored_option)
+    cmake_parse_arguments(ARG "" "FILE" "" ${ARGN})
+    if(NOT ARG_FILE)
+        set(ARG_FILE "CMakeLists.txt")
```
madebr, line 264: Fyi, this never-used  `FILE` argument is a remainer of an attempt to directly use the meson project.

sezero: Revert the changes to PrivateSdlFunctions.cmake ?

madebr: Not needed, just an fyi :)

sezero: Well, I went ahead and reverted it

.github/workflows/main.yml
```diff
@@ -123,11 +129,10 @@ jobs:
           -DSDL2IMAGE_SAMPLES=ON \
           -DSDL2IMAGE_TESTS=ON \
           -DSDL2IMAGE_TESTS_INSTALL=ON \
-          -DSDL2IMAGE_AVIF=${{ !matrix.platform.vendored }} \
+          -DSDL2IMAGE_AVIF=ON \
           -DSDL2IMAGE_JXL=ON \
           -DSDL2IMAGE_TIF=ON \
           -DSDL2IMAGE_WEBP=ON \
-          -DSDL2IMAGE_VENDORED=${{ !!matrix.platform.vendored }} \
```
madebr, line 130: It's defaulted to `ON` on MSVC, but I kept it here to be more explicit.
Also, we test vendored libraries on Linux: see the matrix above.

sezero: How should I change that one? (And feel free to push to my branch.)

madebr: What would you test? Is only testing Windows sufficient?

madebr: Because of this, the MSVC build is veeeeeeery slow.

madebr: Ignore me. Looks like the `-DSDL2IMAGE_VENDORED=` option is duplicated. It's already passed above.

sezero:
One thing I'm curious about is that how this is happening with MSVC:
```
-- Performing Test LINKER_SUPPORTS_NO_UNDEFINED - Success
```

madebr:
> One thing I'm curious about is that how this is happening with MSVC:
> 
> ```
> -- Performing Test LINKER_SUPPORTS_NO_UNDEFINED - Success
> ```

SDL3 has this too. The MSVC linker emits a warning, not an error, when it receives an unknown argument.
[There is a cmake issue tracking this.](https://gitlab.kitware.com/cmake/cmake/-/issues/22023)

sezero:
> > One thing I'm curious about is that how this is happening with MSVC:
> > ```
> > -- Performing Test LINKER_SUPPORTS_NO_UNDEFINED - Success
> > ```
> 
> SDL3 has this too. The MSVC linker emits a warning, not an error, when it receives an unknown argument. [There is a cmake issue tracking this.](https://gitlab.kitware.com/cmake/cmake/-/issues/22023)

Well, we should restrict that to non-windows..  (And also to non-openbsd like SDL2/3..)

sezero:
Anyways, merging this patch. The rest should be subject to new tickets.

PR #398, opened 2023-10-31
use QOI_FREE

baidwwy:
use QOI_FREE


sezero:
Thanks. Also cherry-picked to SDL2 branch.

PR #406, opened 2023-12-14
stb_image.h: update from mainstream.

sezero:
This is prepared against SDL3 branch: If it is accepted, will cherry-pick to SDL2 branch.

If it is accepted, do we want it in the release branch?


slouken:
Let's pull it into SDL2, but not the release branch.

PR #407, opened 2023-12-17
Fix regression (segfault) in LBM image loading [SDL2 branch]

ankith26:
The PR https://github.com/libsdl-org/SDL_image/pull/318 introduced a regression, loading some kinds of `LBM` files now segfaults. 

[Here](https://github.com/libsdl-org/SDL_image/files/13696574/magenta.zip) is a zip that contains one such LBM file that segfaults on 2.8.x but not on earlier versions.

This PR does a simple fix


slouken:
Merged, thanks!

PR #414, opened 2024-01-22
cmake: add SDL3IMAGE_STRICT option to switch between required 3rd party libraries or optional ones

### LINKED ISSUE #413: [cmake] Allow configuring SDL_image with options whose 3rd party dependencies are not available
Currently, CMake configuration fails when a 3rd party dependency is not available.
Let's change this into a soft error and just disable that option.

At the same time, introduce a strict option that fails the build when a dependency could not be fulfilled.

This has been done before in SDL_mixer: https://github.com/libsdl-org/SDL_mixer/commit/2414c440607d9c5c051dee8988699bcb7d97865e

madebr:
A short summary is also printed at the end:
```
-- SDL3_image backends:
-- - enabled:  avif bmp gif jpg jxl lbm pcx png pnm qoi svg tga tif webp xcf xpm xv
-- - disabled: stb imageio wic
```
`jpg` means **a** jpeg backend is available: it can be libjpeg, stb, wic or imageio.

Fixes #413


sezero:
With this patch in, T enabled webp and tif support in cmake file as of 0003e262ea

Also: Can we not do the same in SDL2_image and SDL2_mixer? It's annoying to have one option being disabled by default in cmake while it is enabled by default in autotools.

Issue #416, opened 2024-01-26
longjmp/setjmp clobbered warning in IMG_png.c


madebr:
The following errors appear on ci:
```
 D:/a/SDL_image/SDL_image/src/IMG_png.c:580:16: error: variable 'color_ptr' might be clobbered by 'longjmp' or 'vfork' [-Werror=clobbered]
  580 |     png_colorp color_ptr = NULL;
      |                ^~~~~~~~~
D:/a/SDL_image/SDL_image/src/IMG_png.c:582:18: error: variable 'source' might be clobbered by 'longjmp' or 'vfork' [-Werror=clobbered]
  582 |     SDL_Surface *source = surface;
      |                  ^~~~~~
D:/a/SDL_image/SDL_image/src/IMG_png.c:584:9: error: variable 'png_color_type' might be clobbered by 'longjmp' or 'vfork' [-Werror=clobbered]
  584 |     int png_color_type = PNG_COLOR_TYPE_RGB_ALPHA;
      |         ^~~~~~~~~~~~~~
```
https://github.com/libsdl-org/SDL_image/actions/runs/7661898175/job/20882153620#step:13:29

https://github.com/libsdl-org/SDL_image/blob/abbf3a2addfe8ca9fdd3dd3e1551c14940f0f5e8/src/IMG_png.c#L580-L605

sezero:
In addition to those, gcc-4.9 also warns for `row_pointers` and `surface` in `IMG_LoadPNG_RW` (lines 240 and 247) for me:

```
/tmp/SDL_image/src/IMG_png.c: In function 'IMG_SavePNG_RW_libpng':
/tmp/SDL_image/src/IMG_png.c:580:16: warning: variable 'color_ptr' might be clobbered by 'longjmp' or 'vfork' [-Wclobbered]
     png_colorp color_ptr = NULL;
                ^
/tmp/SDL_image/src/IMG_png.c:584:9: warning: variable 'png_color_type' might be clobbered by 'longjmp' or 'vfork' [-Wclobbered]
     int png_color_type = PNG_COLOR_TYPE_RGB_ALPHA;
         ^
/tmp/SDL_image/src/IMG_png.c: In function 'IMG_LoadPNG_RW':
/tmp/SDL_image/src/IMG_png.c:240:18: warning: variable 'surface' might be clobbered by 'longjmp' or 'vfork' [-Wclobbered]
     SDL_Surface *surface;
                  ^
[...]

PR #417, opened 2024-01-26
IMG_SavePNG / libpng: handle errors at one place, so that libpng/setj…

1bsyl:
IMG_SavePNG / libpng: handle errors at one place, so that libpng/setjump() can also free resources if an error occurs. (see #416)


sezero:
This was integrated in #418 which was just merged. Closing.

PR #431, opened 2024-03-11
JPEG: Handle errors when saving with libjpeg

smcv:
* JPEG: Factor out the middle of IMG_SaveJPG_RW_jpeglib
    
    No functional change intended. We'll need to use setjmp() in this
    function in a subsequent commit, so ensure that its state doesn't
    include any local variables that are used both "above" and "below"
    the stack level at which we will call setjmp().

* JPEG: Add error-recovery when saving with libjpeg
    
    Because we have set up libjpeg to use my_error_exit, we need to call
    setjmp() before the first time it might possibly call longjmp().
    Otherwise, on error we'll do a non-local goto to an uninitialized
    pointer and crash.

---

SDL3 version of #432. Untested (there's no SDL3 in the scout SDK yet) but the only difference is the name of a constant.

### DIFF
--- a/src/IMG_jpg.c
+++ b/src/IMG_jpg.c
@@ -478,15 +478,52 @@ static void jpeg_SDL_RW_dest(j_compress_ptr cinfo, SDL_RWops *ctx)
     dest->pub.free_in_buffer = OUTPUT_BUFFER_SIZE;
 }
 
+struct savejpeg_vars
+{
+    struct jpeg_compress_struct cinfo;
+    struct my_error_mgr jerr;
+};
+
+static int JPEG_SaveJPEG_RW(struct savejpeg_vars *vars, SDL_Surface *jpeg_surface, SDL_RWops *dst, int quality)
+{
+    /* Create a compression structure and load the JPEG header */
+    vars->cinfo.err = lib.jpeg_std_error(&vars->jerr.errmgr);
+    vars->jerr.errmgr.error_exit = my_error_exit;
+    vars->jerr.errmgr.output_message = output_no_message;
+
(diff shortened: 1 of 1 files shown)


smcv:
The Windows CI build says:

```
D:/a/SDL_image/SDL_image/src/IMG_jpg.c:489:18: error: variable 'jpeg_surface' might be clobbered by 'longjmp' or 'vfork' [-Werror=clobbered]
```

smcv:
> Maybe we should separate longjmp handling and the rest like we did for libpng code in [3e42a4f](https://github.com/libsdl-org/SDL_image/commit/3e42a4fb0138f1f2522c43a9e1cda41d24c2a21c) (followed by minor simplification [135bec7](https://github.com/libsdl-org/SDL_image/commit/135bec7c30a00703b3c1a6aab04743ab522262fe)) (also applied to SDL2 branch)

Yes, I think that's going to be necessary. I'm trying to work out how to do that in the most reviewable way.

smcv:
OK, how's this?

sezero:
Looks good, I think. We should do the same to the load function too.

smcv:
> We should do the same to the load function too.

Ideally yes, but I'd prefer that sort of refactoring not to be a blocker for a crash fix.

sezero:
@slouken ?

slouken:
Looks good!

sezero:
We should adapt to SDL2 too and do the same for loader function as well, later.

smcv:
> We should adapt to SDL2 too

That was #432 (which is actually the version I wrote first)

> and do the same for loader function as well, later

#435

Issue #433, opened 2024-03-11
[2.x] Autotools build can detect the wrong libjpeg


smcv:
## To reproduce

```
$ podman pull registry.gitlab.steamos.cloud/steamrt/scout/sdk:beta
$ podman run --rm -it registry.gitlab.steamos.cloud/steamrt/scout/sdk:beta
# cd
# git clone -bSDL2 https://github.com/libsdl-org/SDL_image
# cd SDL_image
  (optionally apply #432 to fix #429)
# autoreconf -fi
# ./configure --enable-tests --disable-stb-image
  (AVIF and JXL libraries are not found, this is OK)
# make
# make check VERBOSE=1
```

Note that in the scout SDK, `/usr/include/jpeglib.h` and `/usr/lib/x86_64-linux-gnu/libjpeg.so` are provided by `libjpeg-turbo8-dev` ([libjpeg-turbo](https://libjpeg-turbo.org/) built to be compatible with the ABI of IJG libjpeg 8), but `/usr/lib/x86_64-linux-gnu/` contains both:

```
/usr/lib/x86_64-linux-gnu/libjpeg.so -> libjpeg.so.8.0.2
/usr/lib/x86_64-linux-gnu/libjpeg.so.8 -> libjpeg.so.8.0.2
```

and

```
/usr/lib/x86_64-linux-gnu/libjpeg.so.62 -> libjpeg.so.62.0.0
```

where the latter is provided by [IJG libjpeg](https://www.ijg.org/) version 6b.

## Expected result

All tests pass.

## Actual result

```
ERROR: 03/11/24 17:37:23: Assert 'Load .../test/sample.jpg (JPEG loading error)': Failed
```

A CMake build from the same sources succeeds.

Looking at the build log, one thing that jumps out at me is:

```
-- dynamic libjpeg -> libjpeg.so.62
```

This means that the libjpeg whose headers we `#include` (libjpeg-turbo, SONAME `libjpeg.so.8`) is not the same as the libjpeg that we dlopen (IJG libjpeg, SONAME `libjpeg.so.62`). libjpeg is sufficiently fault-tolerant that it manages to detect this incompatibility and report a recoverable error, rather than just crashing.

I think this is the root cause for why I experienced #428 and #429, which can only happen in an error-handling code path.

## Workarounds

[...]

PR #434, opened 2024-03-11
Autotools: Allow selecting a specific libjpeg SONAME

smcv:
It is possible to have multiple dynamic shared libraries for libjpeg, for example in the Steam Runtime 1 'scout' SDK, which has development headers and a runtime library for libjpeg-turbo (installed as libjpeg.so.8), but also has the runtime library for IJG libjpeg 6b (installed as libjpeg.so.62). Allow an invocation like

    ./configure --enable-jpg-shared=libjpeg.so.8 ...

to force the use of a specific SONAME.

Resolves: https://github.com/libsdl-org/SDL_image/issues/433

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -239,7 +239,7 @@ AC_ARG_ENABLE([gif], [AS_HELP_STRING([--enable-gif], [support loading GIF images
  [], [enable_gif=yes])
 AC_ARG_ENABLE([jpg], [AS_HELP_STRING([--enable-jpg], [support loading JPG images [default=yes]])],
  [], [enable_jpg=yes])
-AC_ARG_ENABLE([jpg-shared], [AS_HELP_STRING([--enable-jpg-shared], [dynamically load JPG support [default=yes]])],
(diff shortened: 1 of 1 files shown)


sezero:
Can we please expand this to cover other `--enable-XXX-shared` ?

smcv:
See https://github.com/libsdl-org/SDL_image/issues/433#issuecomment-1989302519 for notes on why this is the way it is.

smcv:
> Can we please expand this to cover other `--enable-XXX-shared` ?

For avif, jxl, png and tif: yes if you want, but I'm not really convinced it's worth it just for symmetry. As far as I know, all of these libraries only have one implementation, which is numbered in a straightforward integer sequence (no `62` meaning `6.2`).

For webp: not trivially, we'd have to add a second option, and I'd prefer not to get into that for a stable-branch bug-fix, particularly when Autotools is on its way out (present in 2.x but deleted from 3.x).

sezero:
@slouken ?

smcv:
> For avif, jxl, png and tif: yes if you want, but I'm not really convinced it's worth it just for symmetry.

I added a commit to give these libraries the same treatment, which you can take or not take as you prefer. I'd personally suggest *not* cherry-picking this one to 2.8.x.

PR #442, opened 2024-04-02
cmake: use MACHO* properties to set macho version

madebr:
See https://github.com/libsdl-org/SDL_ttf/pull/349


slouken:
Yeah, we should probably disable ImageIO for the autotools CI. I'm assuming that's what's causing the issue?

madebr:
> Yeah, we should probably disable ImageIO for the autotools CI. I'm assuming that's what's causing the issue?

No, it's only libjxl that's causing issues.
Porting this patch to SDL3_image, I get the same error.

PR #454, opened 2024-06-07
change file format UTF8 -> UTF8 with BOM

jaefunk:
description as title


sezero:
BOM doesn't help with anything, and will likely result in build failures with old compilers. If you lose the BOM, maybe, but not like as it is

PR #482, opened 2024-12-03
Removed IMG_Init() and IMG_Quit()

slouken:


sezero:
avif side looks broken to me:
```diff
--- a/src/IMG_avif.c
+++ b/src/IMG_avif.c
@@ -214,3 +214,3 @@ bool IMG_isAVIF(SDL_IOStream *src)
         /* This might be AVIF, do more thorough checks */
-        if (!IMG_InitAVIF()) {
+        if (IMG_InitAVIF()) {
             avifROData header;
@@ -368,3 +368,3 @@ SDL_Surface *IMG_LoadAVIF_IO(SDL_IOStream *src)
 
-    if ((IMG_InitAVIF()) {
+    if (!IMG_InitAVIF()) {
         return NULL;
```

slouken:
> avif side looks broken to me:

Yes indeed, thank you!

slouken:
@madebr, did I break CI here?

madebr:
> @madebr, did I break CI here?

No, it's the latest libjxl that decodes jxl images in another way, with a delta-color bigger then our tests allow.
https://github.com/libsdl-org/SDL_image/actions/runs/12130547154/job/33821102023?pr=482#step:13:357

madebr:
On the other hand, MSVC also fails which uses our vendored libjxl library. (We vendor an older version)

sezero:
> @madebr, did I break CI here?

```diff
--- a/src/IMG_avif.c
+++ b/src/IMG_avif.c
@@ -118,3 +118,3 @@ static bool IMG_InitAVIF(void)
 
-    return false;
+    return true;
 }
diff --git a/src/IMG_webp.c b/src/IMG_webp.c
index 2541d86..57e76d3 100644
--- a/src/IMG_webp.c
+++ b/src/IMG_webp.c
@@ -100,3 +100,3 @@ static bool IMG_InitWEBP(void)
 
-    return false;
+    return true;
 }
```

icculus (commented):
Getting various drivers to init on demand is good, but how do they clean up without an explicit quit? I saw one of these calls CoCreateInstance during init, for example.

slouken:
> Getting various drivers to init on demand is good, but how do they clean up without an explicit quit? I saw one of these calls CoCreateInstance during init, for example.

They don't, but in practice I don't think unloading the image DLLs is that important these days. The WIC driver is the only one that does CoCreateInstance(), and I think we can potentially just drop that.

PR #487, opened 2024-12-14
Added `--depth 1` flag for cloning external libraries

gxosty:
Not a big change but because my internet speed is always low in my area, I always encounter the problem when external libraries can't be cloned completely (especially libjxl, git failes with RPC call and I assume it's because of country middleboxes...). Everytime, I had to add this flag manually, but I think it won't harm much if it is included in the main repo + it will speed up the download process anyway.

### DIFF
--- a/external/download.sh
+++ b/external/download.sh
@@ -15,7 +15,7 @@ while true; do
     read line; set -- $line
     branch=$3
 
-    if [ -z "${ARGUMENTS}" ]; then
+    if [ -z "$ARGUMENTS" ]; then
         ARGUMENTS="--depth 1"
     fi
 
--- a/external/download.sh
+++ b/external/download.sh
@@ -14,5 +14,5 @@ while true; do
     url=$3
     read line; set -- $line
     branch=$3
-    git clone --filter=blob:none $url $path -b $branch --recursive $ARGUMENTS
+    git clone --filter=blob:none $url $path -b $branch --recursive $ARGUMENTS --depth 1
 done
--- a/external/download.sh
+++ b/external/download.sh
@@ -19,5 +19,5 @@ while true; do
         ARGUMENTS="--depth 1"
     fi
 
-    git clone --filter=blob:none $url $path -b $branch --recursive $ARGUMENTS --depth 1
+    git clone --filter=blob:none $url $path -b $branch --recursive $ARGUMENTS
 done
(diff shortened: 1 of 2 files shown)


madebr:
The shell script already forwards arguments to git. It's used by the release scripts for passing `--depth 1` as well.

How about making it the default, but allowing an override by script arguments?

madebr:
What I mean is add `--depth 1` when ARGUMENTS is empty

gxosty:
alright, makes more sense than passing it directly

gxosty:
Powershell script didn't have ARGUMENTS variable like `download.sh` had, so I added it too

external/download.sh
```diff
@@ -14,5 +14,10 @@ while true; do
     url=$3
     read line; set -- $line
     branch=$3
+
+    if [ -z "${ARGUMENTS}" ]; then
```
slouken, line 18: This is bash syntax. Try this?
```suggestion
    if [ "$ARGUMENTS" = "" ]; then
```

gxosty: oh, sorry. My cmake habit came into action XD. yes that works, I actually wanted to write `if [ -z "$ARGUMENTS" ]; then`. Alright, I will make the changes

slouken:
Merged, thanks!

madebr:
@gxosty 
Looks like there is a bug in the argument handling on PowerShell.
We're seeing this on ci:
https://github.com/libsdl-org/SDL_image/actions/runs/12361676635/job/34499378684#step:6:12

gxosty:
> @gxosty Looks like there is a bug in the argument handling on PowerShell. We're seeing this on ci: https://github.com/libsdl-org/SDL_image/actions/runs/12361676635/job/34499378684#step:6:12

I checked... and I find it weird. Turns out that the default arguments should be split, else `git clone` doesn't understand that `--depth 1` is actually seperated by space. I assume it's the way windows handles arguments passed to processes. This way it works:
```powershell

```

gxosty:
Wait... actually... it works in powershell itself, but when running script, again the same thing

gxosty:
Okay, turns out this way it works

```powershell
[String[]] $Arguments = $args

if ([string]::IsNullOrEmpty($Arguments)) {
    [String[]]$Arguments = "--depth", "1"
}
```

Just changed `$Arguments` type to `String[]`, and assigned default args accordingly.

madebr:
Can you put this in a pr?

PR #488, opened 2024-12-17
Fixed default arguments not being passed properly in external dependencies powershell script

gxosty:
Windows handles arguments passed to processes differently. In PowerShell script, need to explicitly pass array of arguments to `git clone` command. This PR addresses that issue.


slouken:
Merged, thanks!

PR #525, opened 2025-02-18
[SDL2] Update gitignore

FtZPetruska:
This should support the "new" build/test artifacts from autotools and CMake.

Support for CLion and Xcode is also added.


slouken:
Does this need to be done for the other satellite libraries as well? (SDL_mixer, SDL_ttf, etc.?)

FtZPetruska:
> Does this need to be done for the other satellite libraries as well? (SDL_mixer, SDL_ttf, etc.?)

I've given a quick look at these, here are the gitignore that seem like they need updating:
- SDL_mixer on `main` and `SDL2`
- SDL_net on `main`
- SDL_ttf on `SDL2`

I will see about submitting PRs for them.

Issue #524, opened 2025-02-16
IMG_LoadTexture_IO read too much data, far beyond the end of the PNG data


flowCRANE:
Continuation of PNG reading issues reported earlier: #360

If a binary file (or any stream) contains multiple PNG images, one after another, they cannot be loaded correctly using the [IMG_LoadTexture_IO](https://wiki.libsdl.org/SDL3_image/IMG_LoadTexture_IO) function, because the pointer in the stream is moved far beyond the `IEND` chunk of the image.

For example, I have a binary file `sdl.dat` containing two identical PNG images, one after the other. The image size is `10,982` bytes, there are two in the file, so the file size is `21,964` bytes. I open the file and try to load both images into two textures, like this:

```pascal
var
  Stream: PWnc_SDL_Stream;
  Image1: PWnc_SDL_Texture;
  Image2: PWnc_SDL_Texture;
begin
  { initializing SDL, creating window and renderer etc. }

  Stream := SDL_IOFromFile('sdl.dat', 'rb');

  if Stream <> nil then
  begin
    Wnc_LogInfo(['Before:    ', SDL_TellIO(Stream)]);
    Image1 := IMG_LoadTexture_IO(Wnc_RendererGet(), Stream, False);

    if Image1 = nil then
      Wnc_LogError(['Filed to load 1st image: "', SDL_GetError(), '"']);

    Wnc_LogInfo(['After 1st: ', SDL_TellIO(Stream)]);
    Image2 := IMG_LoadTexture_IO(Wnc_RendererGet(), Stream, False);

    if Image2 = nil then
      Wnc_LogError(['Filed to load 2nd image: "', SDL_GetError(), '"']);

    if Image1 <> nil then SDL_DestroyTexture(Image1);
    if Image2 <> nil then SDL_DestroyTexture(Image2);

    if not SDL_CloseIO(Stream) then
      Wnc_LogError(['Failed to close file: "', SDL_GetError(), '"']);
  end
  else
    Wnc_LogError(['Failed to open file: "', SDL_GetError(), '"']);
end.
```
The first image loads correctly, but the second one doesn't. I see an error reading the second image in the console:

```
00  00:00:02.813  INFO     Before:    0
00  00:00:02.815  INFO     After 1st: 11094
00  00:00:02.816  ERROR    Filed to load 2nd image: "Unsupported image format"
00  00:00:02.816  INFO     After 2nd: 11094
```
[...]

PR #526, opened 2025-02-23
fix stbi__parse_png_file() reading too much bytes

### LINKED ISSUE #360: IMG_LoadTexture_RW read too much data, far beyond the end of the PNG data
The `IMG_LoadTexture_RW` function can load a PNG image saved in the stream, but if the stream contains some data after it (because it is, for example, a binary file with a lot of data), then after reading the PNG, the stream pointer is moved much too far, which further data could not be read correctly. The same problem occurs with the `IMG_LoadTextureTyped_RW` function.

The original discussion is here: https://discourse.libsdl.org/t/img-loadtexture-rw-seems-to-load-too-much-data

miniupnp:
just "rewind" the bytes that have been read past the IEND chunk.

should fix #360 and #524

### DIFF
--- a/src/stb_image.h
+++ b/src/stb_image.h
@@ -5395,6 +5395,10 @@ static int stbi__parse_png_file(stbi__png *z, int scan, int req_comp, unsigned i
             STBI_FREE(z->expanded); z->expanded = NULL;
             // end of PNG chunk, read and skip CRC
             stbi__get32be(s);
+            if (s->io.skip && s->img_buffer_end > s->img_buffer) {
+               // rewind the additional bytes that have been read to the buffer
+               (s->io.skip)(s->io_user_data, (int)(s->img_buffer - s->img_buffer_end));
+            }
             return 1;
          }
 


miniupnp:
@flowCRANE could you check if that fixes your issue ?

I added a couple of bytes at the end of a file and checked using ftell() that is is indeed rewinding the additional bytes ;)

I guess I should submit the patch to https://github.com/nothings/stb if it works for you

flowCRANE:
If it works for you, it will work for me. You can test it on my sample file — [test.files.zip](https://github.com/user-attachments/files/18943763/test.files.zip)

If you are able to open `sdl.bin` file and read two images from the stream successively, without any errors, in this way:

```pascal
Stream := SDL_IOFromFile('sdl.bin', 'rb');

Image1 := IMG_LoadTexture_IO(Renderer, Stream, False);
Image2 := IMG_LoadTexture_IO(Renderer, Stream, False);
```
then it means that your fix is correct and solves the problem. And thanks so much for looking into this topic and finding the cause of the problems!

miniupnp:
Some Pascal :)

@flowCRANE So you are not able to test the patch ?

miniupnp:
With the following code, img2 is null without the patch and is not null with the patch.
So I guess it works.
```c
{
	SDL_IOStream *src = SDL_IOFromFile("/home/nanard/code/git/SDL_image/sdl.dat", "rb");
	SDL_Texture *img1 = IMG_LoadTexture_IO(renderer, src, false);
	SDL_Texture *img2 = IMG_LoadTexture_IO(renderer, src, true);
        printf("src=%p img1=%p img2=%p\n", src, img1, img2);
}
```

miniupnp:
https://github.com/nothings/stb/pull/1751

slouken:
Merged, thanks!

flowCRANE:
@miniupnp: yeah, I'm in Free Pascal's camp and have no thought of betraying the colors. ;)

The main idea is that the data stream can be constructed in such a way that multiple PNG images can be written one after the other and that they can be loaded one after the other, like in the example above. This will mean that when the image is read, the pointer is in the correct place.

I haven't done extensive testing, so I don't know if the same problem exists for other formats. But I'll look into it and let you know if I find anything. Thanks again for the fix!

sezero:
> Merged, thanks!

Add to SDL2 branch too.

And also to SDL3 main repo?

slouken:
Yep, done

sezero:
Do we want https://github.com/miniupnp/stb/pull/1 ?  Maybe not here (because `stbi__start_mem()` is unused in SDL_image), but in SDL itself?

slouken:
> Do we want [miniupnp/stb#1](https://github.com/miniupnp/stb/pull/1) ? Maybe not here (because `stbi__start_mem()` is unused in SDL_image), but in SDL itself?

Yes.

PR #528, opened 2025-02-26
cmake: don't search libXXX if SDLIMAGE_XXX_SHARED is ON

podsvirov:


madebr (approved):
Thanks!

While you're at it, can you also add `AND NOT SDLIMAGE_TIF_SHARED` for libtiff?
(In the same way as we did for libjpeg)

podsvirov:
> While you're at it, can you also add `AND NOT SDLIMAGE_TIF_SHARED` for libtiff?

Done.

madebr:
Merged!
Thanks for the pr!

PR #531, opened 2025-03-03
Handle absolute include and lib paths in cmake

marcin-serwin:
If the `CMAKE_INSTALL_INCLUDEDIR` or `CMAKE_INSTALL_LIBDIR` are absolute it will result in a broken paths in the resulting `.pc` file.

This patch is based on the https://github.com/libsdl-org/SDL/pull/10826 PR from the SDL repo which solves identical issue.


slouken:
@madebr, can you commit and merge to the appropriate branches and other satellite libraries?

madebr:
> @madebr, can you commit and merge to the appropriate branches and other satellite libraries?

Done :)

PR #539, opened 2025-04-02
cmake: configure external subprojects in different directory

madebr:
(This avoids in-tree builds of external projects when doing in-tree SDL builds)

Also fixes #538

PR #540, opened 2025-04-02
Disable sjpeg and openexr support of vendored libjxl

madebr:
(cherry picked from commit 5df4a32e14a3ed46ef0a4f33132c831331ef0b8f)

sdl2 backport of #539


sezero:
Why is it still failing despite `...3.10` ?

madebr:
> Why is it still failing despite `...3.10` ?

It's doing:
```cmake
# Default policy is from 2.8.9
cmake_policy(VERSION 2.8.9)
```

madebr:
Btw, when setting the CMake policy compatibility version, it's okay to set it as high as possible.

CMake then knows the script is compatible with this version, and we won't have compatibility issues for many years to come.

PR #542, opened 2025-04-08
rename local vars 'pixel' to 'pixelvalue' or 'pixels'

sezero:
Reference issue: https://github.com/libsdl-org/SDL/issues/12749.

(CC @barracuda156 )


barracuda156:
I will rebuild with this one added now.

slouken (approved):
Should we upstream the stb_image.h change?

Issue #435, opened 2024-03-11
Factor out the parts of IMG_LoadJPG_RW where locals can be clobbered by setjmp()/longjmp()


smcv:
Ideally the libjpeg loading path should have an equivalent of #418 and https://github.com/libsdl-org/SDL_image/commit/f98f0bcd9b35ba6c2f77db042cc056d245be81c6.

(Originally requested by @sezero in https://github.com/libsdl-org/SDL_image/pull/431#issuecomment-1989322714)

PR #544, opened 2025-04-09
factor out parts of IMG_LoadJPG where locals can be clobbered by longjmp (SDL2 version)

sezero:
Closes: https://github.com/libsdl-org/SDL_image/issues/435

CC: @smcv


smcv:
Same comments as in #545

sezero:
PING @slouken

sezero:
This is in now. If there are any issues, please open a new ticket for it.

PR #545, opened 2025-04-09
factor out parts of IMG_LoadJPG where locals can be clobbered by longjmp (SDL3 version)

sezero:
Closes: https://github.com/libsdl-org/SDL_image/issues/435

CC: @smcv

### DIFF
--- a/src/IMG_jpg.c
+++ b/src/IMG_jpg.c
@@ -124,6 +124,7 @@ static bool IMG_InitJPG(void)
 
     return true;
 }
+
 #if 0
 void IMG_QuitJPG(void)
 {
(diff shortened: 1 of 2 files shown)


src/IMG_jpg.c
```diff
@@ -343,89 +343,103 @@ static void output_no_message(j_common_ptr cinfo)
...
-        surface = SDL_CreateSurface(cinfo.output_width, cinfo.output_height, SDL_PIXELFORMAT_RGB24);
+        vars->surface = SDL_CreateSurface(vars->cinfo.output_width, vars->cinfo.output_height, SDL_PIXELFORMAT_RGB24);
     }
 
-    if ( surface == NULL ) {
-        lib.jpeg_destroy_decompress(&cinfo);
-        SDL_SeekIO(src, start, SDL_IO_SEEK_SET);
+    if (!vars->surface) {
+        lib.jpeg_destroy_decompress(&vars->cinfo);
         SDL_SetError("Out of memory");
```
smcv, line 406: Not a regression, but on failure didn't `SDL_CreateSurface()` already set the error indicator to something more specific?

src/IMG_jpg.c
```diff
@@ -343,89 +343,103 @@ static void output_no_message(j_common_ptr cinfo)
...
+    if (!IMG_InitJPG()) {
+        return NULL;
+    }
+
+    start = SDL_TellIO(src);
+    SDL_zero(vars);
+    LIBJPEG_LoadJPG_IO(src, &vars);
+
+    if (!vars.surface) {
+        SDL_SeekIO(src, start, SDL_IO_SEEK_SET);
```
smcv, line 439: I think this could clobber the error indicator if the seek fails, so it could need the equivalent of the `int saved_errno = errno` pattern. Perhaps something like this (untested):

```suggestion
        char *saved_error = SDL_strdup(SDL_GetError());

        SDL_SeekIO(src, start, SDL_IO_SEEK_SET);
        if (saved_error) {
            SDL_SetError("%s", saved_error);
        }
        else {
            SDL_OutOfMemory();
        }
        SDL_free(saved_error);
```

src/IMG_jpg.c
```diff
@@ -343,89 +343,103 @@ static void output_no_message(j_common_ptr cinfo)
     (void)cinfo;
 }
 
-/* Load a JPEG type image from an SDL datasource */
-SDL_Surface *IMG_LoadJPG_IO(SDL_IOStream *src)
-{
-    Sint64 start;
+#ifdef _MSC_VER
+#pragma warning(disable:4611)   /* warning C4611: interaction between '_setjmp' and C++ object destruction is non-portable */
+#endif
```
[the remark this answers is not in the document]
sezero: This one I hadn't tested, so I had kept it. But now I just removed it both here and in IMG_png.

src/IMG_jpg.c
sezero: Can you please make this (and the other above) into a new ticket?

src/IMG_jpg.c
sezero: Possibly, but all other loaders do the same, so I didn't change the behavior in this patch.

Maybe a subject for another patch.

sezero:
PING @slouken

sezero:
@smcv :  Is this OK with you?  (Also its SDL2 counterpart.) The error message stuff you noted requires another go at the full source tree, as far as I can see.

sezero:
Well, I did went ahead and sanitized error setting _a bit_ for png and jpg loaders, both here and in the SDL2 version as well. Please review.

sezero:
This is in now. If there are any issues, please open a new ticket for it.

PR #548, opened 2025-04-12
cmake: disable SDLIMAGE_(AVIF|JPG|PNG)_SAVE when backends are not available

### LINKED ISSUE #547: cmake: avif save option not disabled when no avif backend is found
As title says: That apparently leads to test failure..

And I guess same issue is present with png and jpg, too.

CC @madebr

madebr:
Fixes #547

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -569,6 +569,9 @@ if(SDLIMAGE_AVIF)
         else()
             target_link_libraries(${sdl3_image_target_name} PRIVATE avif)
         endif()
+    else()
+        # Variable is used by test suite
+        set(SDLIMAGE_AVIF_SAVE OFF)
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


sezero:
Tested and it works for me.

Issue _possibly_ present in SDL2 version, too.

sezero:
Looking briefly at png and jpeg: even though external libs aren't found, tinyjpeg and miniz are there, so jpeg and png aren't restricted by backend presence? (If so, sdl2 version isn't affected at all.)

sezero:
> Looking briefly at png and jpeg: even though external libs aren't found, tinyjpeg and miniz are there, so jpeg and png aren't restricted by backend presence? (If so, sdl2 version isn't affected at all.)

OK, looked closely and tried: IMG_png.c and IMG_jpg.c are messy about requirements when deciding to enable save option, even manually adding save defines among cflags fails to build the miniz and tinyjpeg code when libpng/libjpeg aren't available and stb is disabled. So, I guess your patch is OK as it is.

PR #551, opened 2025-04-20
IMG_stb: Fix a memory leak for PNG.

iyzsong:
Hello, the `SDL_PREALLOC` flag should be cleared before `SDL_ConvertSurfaceFormat`, so that the pixels would be freed.


slouken:
Merged, thanks!

sezero:
Is SDL3_image not affected?

slouken:
> Is SDL3_image not affected?

Nope.

Issue #470, opened 2024-09-29
`ApplicationServices` framework is not available on `iOS`?


st235:
It appears that the `ApplicationServices` framework is linked across the entire `Apple` ecosystem in the [CMakeLists.txt](https://github.com/libsdl-org/SDL_image/blob/main/CMakeLists.txt#L371). When building for `iOS` both `IOS` and `APPLE` flags are defined, therefore `ApplicationServices` is linked to the target. 

However, I believe it is not available on `iOS`, as generating an `Xcode` project with `SDL3_image` linked for iOS results in a build failure.

![Screenshot 2024-09-29 at 17 24 19](https://github.com/user-attachments/assets/29c9149d-9948-4cd2-9aaf-1fc953a071d6)

I am wondering whether I am doing anything wrong or this is really a bug in the build script?

This is my `CMakeLists.txt` build script if of any use (some details are lifted for the sake of simplicity):

```cmake
FetchContent_Declare(
    SDL_IMAGE
    GIT_REPOSITORY https://github.com/libsdl-org/SDL_image.git
    GIT_TAG main
    GIT_SHALLOW TRUE
    GIT_PROGRESS TRUE
)

FetchContent_MakeAvailable(SDL_IMAGE)

add_library(core INTERFACE)
target_sources(core INTERFACE
  src/app.h
  src/app.cpp
)
target_include_directories(core INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
target_link_libraries(core INTERFACE SDL3::SDL3 SDL3_image::SDL3_image)
```

And this is the build command I use

```bash
cmake -B out/ios -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES="arm64;i386;x86_64"
```

P.S.: the declaration of `SDLIMAGE_BACKEND_IMAGEIO` slightly hints that this is only related to `Mac OS X`.

```cmake
cmake_dependent_option(SDLIMAGE_BACKEND_IMAGEIO "Use native Mac OS X frameworks for loading images" ON APPLE OFF)
```

P.P.S.: [this stack overflow thread](https://stackoverflow.com/questions/9393295/applicationservices-framework-not-found-error-in-ios) also suggests that `ApplicationServices` are not available in `iOS`.

[...]

PR #553, opened 2025-05-04
ISSUE-470: add CoreGraphics, ImageIO, and MobileCoreServices required from `IMG_ImageIO.mm` for `iOS`-specific builds

st235:
The `ApplicationServices` framework, which includes `ImageIO`-specific definitions, is not available on `iOS`. It was removed in [Issue 470](https://github.com/libsdl-org/SDL_image/issues/470), and no replacement is currently linked for `iOS` builds.

As a result, without linking the appropriate `iOS` frameworks, the [`ImageIO`](https://developer.apple.com/documentation/imageio)  backend won't function on `iOS`.
However, [`ImageIO`](https://developer.apple.com/documentation/imageio) itself has been available on `iOS` since version `4.0`.

Screenshot 2025-05-04 at 12 00 20

A temporary workaround is to disable the `ImageIO` backend for `iOS` builds, although there don’t appear to be any technical reasons preventing its use:

```cmake
set(SDL2IMAGE_BACKEND_IMAGEIO OFF)
```

To enable `ImageIO` support on `iOS`, the build should link against the following frameworks:

* `Core Graphics` - for `CGContext` calls like `CGBitmapContextCreate`
* `ImageIO` - for `CGImageSource` functions such as `CGImageSourceCreateImageAtIndex`
* `MobileCoreServices` - for definitions like `UTTypeGIF`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -383,6 +383,12 @@ if(APPLE)
         set(SDLIMAGE_IMAGEIO_ENABLED TRUE)
         if(CMAKE_SYSTEM_NAME MATCHES ".*(Darwin|MacOS).*")
             target_link_libraries(${sdl3_image_target_name} PRIVATE -Wl,-framework,ApplicationServices)
+        elseif(IOS)
+            target_link_libraries(${sdl3_image_target_name} PRIVATE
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #556, opened 2025-05-19
improve INTRO-visualstudio.md with more detailed Visual Studio steps

Atkclf:
The original guide was missing the steps for adding the SDL3 include directories and the SDL3 reference in the SDL3_image project. I have added these steps.


slouken:
Merged, thanks!

PR #559, opened 2025-06-03
Improve palette handling in STB backend

ankith26:
fixes https://github.com/libsdl-org/sdl2-compat/issues/492

While I was investigating that issue I realised that this must actually be fixed on the `SDL_image` end and not `sdl2-compat` end as I initially assumed.

This PR makes the stbimage backend path more like the libpng backend:
- If a colorkey is set along with a palette, the colorkey entry shouldn't have redundant pixel alpha of 0 as that messes up the `SDL_MapRGB` logic.
- The stbimage backend always gives 256 entries in the palette, even if less entries are required. It seems like stb image does not expose this information in its API. I have put a bit of a hacky solution to this issue. It's certainly not very efficient but is the easiest approach I could think of.

### DIFF
--- a/src/IMG_stb.c
+++ b/src/IMG_stb.c
@@ -191,6 +191,8 @@ SDL_Surface *IMG_LoadSTB_RW(SDL_RWops *src)
                 SDL_FreeSurface(surface);
                 surface = converted;
             } else if (has_colorkey) {
+                /* remove redundant pixel alpha before setting colorkey */
+                palette->colors[colorkey_index].a = SDL_ALPHA_OPAQUE;
                 SDL_SetColorKey(surface, SDL_TRUE, colorkey_index);
             }
         }


ankith26:
If this PR is accepted I can remake this PR against main/SDL3

slouken:
Iterating over all the pixels to find the highest index is relatively expensive. Is there a change to stb_image that we can make, to retain the correct palette size?

ankith26:
> Is there a change to stb_image that we can make, to retain the correct palette size?

I'm not aware how to do this rn. My assumption here is that its very unlikely to have an 8 bit image that is also very large, so I think the added complexity should be okay-ish.

src/IMG_stb.c
```diff
@@ -191,6 +203,8 @@ SDL_Surface *IMG_LoadSTB_RW(SDL_RWops *src)
                 SDL_FreeSurface(surface);
                 surface = converted;
             } else if (has_colorkey) {
+                /* remove redundant pixel alpha before setting colorkey */
+                palette->colors[colorkey_index].a = SDL_ALPHA_OPAQUE;
```
ankith26, line 195: If the palette reduction part of this PR is not acceptable to approve, should I make a separate PR with just this one line fix? IMO this is a more critical fix, without this fix `SDL_MapRGB` is completely broken on such kind of surfaces.

slouken: Yes please.

slouken: Can you also provide a quick example program that demonstrates why this is necessary?

ankith26: The original issue report https://github.com/libsdl-org/sdl2-compat/issues/492 has an example program along with a more detailed explanation of the issue.

UPDATE: I have updated this PR to only do this fix, and removed the other changes.

slouken:
Merged, thanks!

PR #560, opened 2025-06-03
Save non-alpha pngs as RGB not RGBA, handle arbitrary surface pitch properly (miniz backend)

ankith26:
This PR is an attempt to make the behaviour of SavePNG more consistent across backends (miniz and libpng). With this PR, images that don't need alpha will now be stored as RGB instead of RGBA.

### DIFF
--- a/src/IMG_png.c
+++ b/src/IMG_png.c
@@ -812,8 +812,7 @@ static int IMG_SavePNG_RW_miniz(SDL_Surface *surface, SDL_RWops *dst)
                 png = tdefl_write_image_to_png_file_in_memory(tightened_pixels, w, h, bpp, tightened_pitch, &size);
                 SDL_free(tightened_pixels);
             }
-        }
-        else {
+        } else {
             png = tdefl_write_image_to_png_file_in_memory(pixels, w, h, bpp, pitch, &size);
         }
     }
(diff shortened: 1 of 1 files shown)


ankith26:
As a side goal of this PR, the code is now more robust and handles arbitrary pitches. It seems like miniz always expects a tight pitch (and generates corrupted PNGs if this isn't the case), but the input surface or the surface output of `SDL_ConvertSurfaceFormat` need not always give out a surface with tight pitch (especially in the RGB case where it always seemingly aligns the pitch to a multiple of 4)

src/IMG_png.c
```diff
@@ -766,15 +766,61 @@ static int IMG_SavePNG_RW_miniz(SDL_Surface *surface, SDL_RWops *dst)
...
+            Uint8 *tightened_pixels = SDL_malloc(tightened_pitch * h);
+            if (tightened_pixels) {
+                for (int y = 0; y < h; y++) {
+                    SDL_memcpy(tightened_pixels + y * tightened_pitch, pixels + y * pitch, tightened_pitch);
+                }
+                png = tdefl_write_image_to_png_file_in_memory(tightened_pixels, w, h, bpp, tightened_pitch, &size);
+                SDL_free(tightened_pixels);
+            }
+        }
+        else {
```
slouken:
```suggestion
        } else {
```

slouken:
SDL_PIXELFORMAT_RGB24 is the slowest format to convert to. Does the libpng code actually do this?

ankith26:
Yes, the libpng backend does convert to `SDL_PIXELFORMAT_RGB24` for non-alpha cases.

slouken:
I think this change would be okay in SDL3, but we shouldn't change SDL2 behavior at this point. Feel free to open a new PR against main if it's still relevant.

Issue #570, opened 2025-07-24
IMG_Load_IO Fails to Load Certain WebP Files


LordOfXen:
There seems to be an API inconsistency with how GIF and WebP images are loaded with SDL_image.
The IMG_Load_IO successfully loads an animated gif image's first frame into a SDL_Surface, however, it fails to do so the same with an animated WebP image file (I've tried two animated WebP images and both fails).

If IMG_Load_IO is supposed to load only single-frame images, why does it success with a GIF file and not with a WebP file? Considering SDL, it should either fail for both or success for both.

This is the test code in case you want to test it yourself:
```cpp
// SDL3ImageTestbed.cpp : This file contains the 'main' function. Program execution begins and ends there.
//

#include <iostream>
#include "SDL3/SDL.h"
#include "SDL3_image/SDL_image.h"

int main()
{
    if (!SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS)) {
        std::cerr << "Failed to initialize SDL: " << SDL_GetError() << std::endl;
        return 1;
	}

    SDL_IOStream* stream = SDL_IOFromFile("test2.webp", "r");
    if (!stream)
    {
        std::cerr << "Failed to open file: " << SDL_GetError() << std::endl;
        SDL_Quit();
        return 1;
	}

    IMG_Animation* animation = IMG_LoadAnimation_IO(stream, false);
    if (!animation) {
        SDL_CloseIO(stream);
		stream = nullptr;

        std::cerr << "Failed to load animation: " << SDL_GetError() << std::endl;
        SDL_Quit();
        return 1;
	}

    printf("Animation loaded successfully:\nFrame Count: %i\n", animation->count);

	// Seek the stream to the beginning to be able to read the first frame with IMG_Load_IO.
    if (SDL_SeekIO(stream, 0, SDL_IO_SEEK_SET) != 0)
    {
        SDL_CloseIO(stream);
        stream = nullptr;

        std::cerr << "Failed to load first frame/seeking the IO failed: " << SDL_GetError() << std::endl;
        SDL_Quit();
        return 1;
    }

    SDL_Surface* imgFirstFrame = IMG_Load_IO(stream, false);
    if (!imgFirstFrame)
    {
        SDL_CloseIO(stream);
        stream = nullptr;

        std::cerr << "Failed to load first frame: " << SDL_GetError() << std::endl;
        SDL_Quit();
        return 1;
	}

[...]

PR #572, opened 2025-07-28
Update IMG_webp.c

LordOfXen:
## What it does?
Fix API inconsistency with allowing animated WebP images to be used with IMG_LoadWEBP_IO to extract a single frame, just like GIF.

This basically calls IMG_LoadWEBPAnimation_IO to get the frame list, then extracts the first frame to return within IMG_LoadWEBP_IO, evaluating to API consistency with how SDL handles GIF images with IMG_Load.

Some additions, like combining the pre-defined error code with the error code from SDL_BlitSurface, might be added if necessary.
Please ask me to make changes if required.

## What was the Issue?
This resolves https://github.com/libsdl-org/SDL_image/issues/570 directly with the suggested changes.

### DIFF
--- a/src/IMG_webp.c
+++ b/src/IMG_webp.c
@@ -221,7 +221,7 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
 
     // Special casing for animated WebP images to extract a single frame.
     if (features.has_animation) {
-        if (SDL_SeekIO(src, start, SDL_IO_SEEK_SET) != 0) {
+        if (SDL_SeekIO(src, start, SDL_IO_SEEK_SET) < 0) {
             error = "Failed to seek IO to read animated WebP";
             goto error;
         } else {
--- a/src/IMG_webp.c
+++ b/src/IMG_webp.c
@@ -357,7 +357,7 @@ static IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int m
     anim->w = features.width;
     anim->h = features.height;
     uint32_t fc = lib.WebPDemuxGetI(demuxer, WEBP_FF_FRAME_COUNT);
-    anim->count = maxFrames > 0 ? SDL_min(maxFrames, fc) : fc;
+    anim->count = maxFrames > 0 ? SDL_min((unsigned)maxFrames, fc) : fc;
     anim->frames = (SDL_Surface **)SDL_calloc(anim->count, sizeof(*anim->frames));
     anim->delays = (int *)SDL_calloc(anim->count, sizeof(*anim->delays));
     if (!anim->frames || !anim->delays) {
@@ -387,10 +387,9 @@ static IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int m
 
     SDL_zero(iter);
 
-    int frame_idx = 0;
     if (lib.WebPDemuxGetFrame(demuxer, 1, &iter)) {
         do {
-            frame_idx = (iter.frame_num - 1);
+            int frame_idx = (iter.frame_num - 1);
             if (frame_idx < 0 || frame_idx >= anim->count) {
                 continue;
             }
@@ -427,7 +426,7 @@ static IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int m
             anim->delays[frame_idx] = iter.duration;
             dispose_method = iter.dispose_method;
 
-        } while ((maxFrames > 0 ? frame_idx + 1 < maxFrames : true) && lib.WebPDemuxNextFrame(&iter));
+        } while (lib.WebPDemuxNextFrame(&iter));
 
         lib.WebPDemuxReleaseIterator(&iter);
     }
--- a/src/IMG_webp.c
+++ b/src/IMG_webp.c
@@ -316,10 +316,6 @@ IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int maxFrame
         return NULL;
     }
 
-    if (maxFrames < 0) {
-        maxFrames = 0; /* no limit */
-    }
-
     start = SDL_TellIO(src);
 
     if (!IMG_InitWEBP()) {
--- a/src/IMG_webp.c
+++ b/src/IMG_webp.c
@@ -231,10 +231,41 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
         goto error;
     }
 
-    if (features.has_alpha) {
-        ret = lib.WebPDecodeRGBAInto(raw_data, raw_data_size, (uint8_t *)surface->pixels, surface->pitch * surface->h,  surface->pitch);
+    // Special casing for animated WebP images to extract a single frame.
+    if (features.has_animation) {
+        if (SDL_SeekIO(src, 0, SDL_IO_SEEK_SET) != 0) {
+            error = "Failed to seek IO to read animated WebP";
+            goto error;
+        } else {
+            IMG_Animation *animation = IMG_LoadWEBPAnimation_IO(src);
+            if (animation && animation->count > 0) {
+                SDL_Surface *surf = animation->frames[0];
+                if (surf) {
+                    if (SDL_BlitSurface(surf, NULL, surface, NULL)) {
+                        IMG_FreeAnimation(animation);
+                    } else {
+                        error = "Failed to blit first frame of animated WebP";
+                        // TODO: Maybe get the error from SDL_BlitSurface with SDL_GetError(), then combine it with the error message so it contains all required information?
+                        IMG_FreeAnimation(animation);
+                        goto error;
+                    }
+                } else {
+                    error = "Failed to load first frame of animated WebP";
+                    IMG_FreeAnimation(animation);
+                    goto error;
+                }
+            } else {
+                IMG_FreeAnimation(animation);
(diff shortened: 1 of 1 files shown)


src/IMG_webp.c
```diff
@@ -231,10 +231,41 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
...
+    if (features.has_animation) {
+        if (SDL_SeekIO(src, 0, SDL_IO_SEEK_SET) != 0) {
+            error = "Failed to seek IO to read animated WebP";
+            goto error;
+        } else {
+            IMG_Animation *animation = IMG_LoadWEBPAnimation_IO(src);
+            if (animation && animation->count > 0) {
+                SDL_Surface *surf = animation->frames[0];
+                if (surf) {
+                    if (SDL_BlitSurface(surf, NULL, surface, NULL)) {
```
slouken: Instead of blitting, you might be able to just retain a reference to the original surface, e.g.:
```c
SDL_DestroySurface(surface);
surface = surf;
++surface->refcount;
IMG_FreeAnimation(animation);
```

slouken: (or skip allocating `surface` above in this case...)

LordOfXen: Yeah sure, let me take a look.

LordOfXen:
Now it uses ++ref instead of SDL_BlitSurface. Have tested as well, works as intended.

src/IMG_webp.c
```diff
@@ -219,10 +219,38 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
         goto error;
     }
 
+    // Special casing for animated WebP images to extract a single frame.
+    if (features.has_animation) {
+        if (SDL_SeekIO(src, 0, SDL_IO_SEEK_SET) != 0) {
```
slouken:
```suggestion
        if (SDL_SeekIO(src, start, SDL_IO_SEEK_SET) != 0) {
```

src/IMG_webp.c
```diff
@@ -219,10 +219,38 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
...
+        } else {
+            IMG_Animation *animation = IMG_LoadWEBPAnimation_IO(src);
+            if (animation && animation->count > 0) {
+                SDL_Surface *surf = animation->frames[0];
+                if (surf) {
+                    ++surf->refcount;
+                    if (raw_data) {
+                        SDL_free(raw_data);
+                    }
+                    return surf;
```
slouken, line 239:
```suggestion
                    IMG_FreeAnimation(animation);
                    return surf;
```

LordOfXen:
I also have a recommendation. This will perform a slow operation for big animations, for example, it will extract 500 frames, just to get the first one, then destroy the rest 499. Those 499 frames will take time just to be garbaged.

Instead, maybe we can move IMG_LoadWEBPAnimation_IO's body into an internal function with max frames parameter, use a default value like 0 or -1 to allow all frames to be loaded or a valid value to load "that" many of frames, then use that internal function here to give 1 so it does not load all those 500 frames and instead it just loads the first one and returns.

This would be much faster for SDL, what do you think?

src/IMG_webp.c
```diff
@@ -219,10 +219,39 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
         goto error;
     }
 
+    // Special casing for animated WebP images to extract a single frame.
+    if (features.has_animation) {
+        if (SDL_SeekIO(src, start, SDL_IO_SEEK_SET) != 0) {
```
sezero: Change that to `< 0` -- otherwise it won't do what you expect if start isn't 0.

LordOfXen: Oh right, will do so now. Please also check my recommendation, I wonder your thoughts.

slouken:
> I also have a recommendation. This will perform a slow operation for big animations, for example, it will extract 500 frames, just to get the first one, then destroy the rest 499. Those 499 frames will take time just to be garbaged.
> 
> Instead, maybe we can move IMG_LoadWEBPAnimation_IO's body into an internal function with max frames parameter, use a default value like 0 or -1 to allow all frames to be loaded or a valid value to load "that" many of frames, then use that internal function here to give 1 so it does not load all those 500 frames and instead it just loads the first one and returns.
> 
> This would be much faster for SDL, what do you think?

This seems like a good idea.

LordOfXen:
Please do not merge this now then, let me implement that internal version first to see how it goes. Then we can decide between both before merging.

LordOfXen:
I have implemented the mentioned improvement, please take a look.

src/IMG_webp.c
```diff
@@ -166,6 +166,8 @@ bool IMG_isWEBP(SDL_IOStream *src)
     return webp_getinfo(src, NULL);
 }
 
+IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int maxFrames);
```
slouken:
```suggestion
static IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int maxFrames);
```

src/IMG_webp.c
```diff
@@ -266,7 +296,7 @@ SDL_Surface *IMG_LoadWEBP_IO(SDL_IOStream *src)
     return NULL;
 }
 
-IMG_Animation *IMG_LoadWEBPAnimation_IO(SDL_IOStream *src)
+IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int maxFrames)
```
slouken:
```suggestion
static IMG_Animation *IMG_LoadWEBPAnimation_IO_Internal(SDL_IOStream *src, int maxFrames)
```

src/IMG_webp.c
```diff
@@ -286,6 +316,10 @@ IMG_Animation *IMG_LoadWEBPAnimation_IO(SDL_IOStream *src)
         return NULL;
     }
 
+    if (maxFrames < 0) {
+        maxFrames = 0; /* no limit */
+    }
+
```
slouken:
```suggestion
```
This is an internal function and we never call this with a negative number.

src/IMG_webp.c
```diff
@@ -355,9 +390,11 @@ IMG_Animation *IMG_LoadWEBPAnimation_IO(SDL_IOStream *src)
...
 
     SDL_zero(iter);
+
+    int frame_idx = 0;
     if (lib.WebPDemuxGetFrame(demuxer, 1, &iter)) {
         do {
-            int frame_idx = (iter.frame_num - 1);
+            frame_idx = (iter.frame_num - 1);
             if (frame_idx < 0 || frame_idx >= anim->count) {
                 continue;
```
slouken: I think this continue means that we'll skip frames outside our desired range, since we set `anim->count` above, right? If so, we can remove the awkward while condition below.

LordOfXen: Exactly.

LordOfXen: I will change that, a minute.

LordOfXen: How about getting rid of the while condition completely? We can instead use that `lib.WebPDemuxNextFrame(&iter)` in an if condition as well and rather use a for too, to iterate as many times as `anim->count` states.

slouken: I believe the spec allows frames out of order.

LordOfXen: I saw this: 
> in the order they appear in the WebP file

But I am not sure, the test "for" loop I implemented seem to be working though but need to make sure how WebPDemuxNextFrame returns.

LordOfXen: Here I found this documentation:

> Note: dmux must persist for the lifetime of iter.
> 
> WebPDemuxNextFrame(), WebPDemuxPrevFrame()
> Sets iter->fragment to point to the next (iter->frame_num + 1) or previous (iter->frame_num - 1) frame. These functions do not loop.

According to this we can actually use a for loop, can't we?

Source: https://developers.google.com/speed/webp/docs/container-api

src/IMG_webp.c
```diff
@@ -326,7 +356,8 @@ IMG_Animation *IMG_LoadWEBPAnimation_IO(SDL_IOStream *src)
     }
     anim->w = features.width;
     anim->h = features.height;
-    anim->count = lib.WebPDemuxGetI(demuxer, WEBP_FF_FRAME_COUNT);
+    uint32_t fc = lib.WebPDemuxGetI(demuxer, WEBP_FF_FRAME_COUNT);
+    anim->count = maxFrames > 0 ? SDL_min(maxFrames, fc) : fc;
```
sezero: Casting maxFrames to (unsigned) in SDL_min() should get rid of the build errors

src/IMG_webp.c
slouken: Please don't change the loop.

LordOfXen: Alright.

LordOfXen: You just want me to get rid of this:  `(maxFrames > 0 ? frame_idx + 1 < maxFrames : true) &&`

Correct?

src/IMG_webp.c
LordOfXen: Alrighty! I will just replace maxFrames with (unsigned)maxFrames in SDL_min.

LordOfXen:
Got rid of the unnecessary check in the while condition (moved frame_idx back inside), and added unsigned to prevent build errors.

slouken:
Looks good, thanks!

PR #574, opened 2025-07-30
Seek stream to the beginning if error in IMG_SaveWEBP_IO

LordOfXen:
IMG_LoadWEBP_IO and IMG_LoadWEBPAnimation_IO_Internal get the position at the beginning and set it back to the stream if the operation fails. Following this standard, the same was added to our new save function, so it behaves the same.


slouken:
Merged, thanks!

PR #575, opened 2025-07-30
Add Functionality to Check if WebP is Animated (Declarations & Definitions) 

LordOfXen:
2 functions to check whether the provided WebP stream or file is animated or not. This is kinda mandatory, or you can't know where to push the WebP image, or have to load the entire file to check if it's animated or not.

### DIFF
--- a/include/SDL3_image/SDL_image.h
+++ b/include/SDL3_image/SDL_image.h
@@ -2027,6 +2027,41 @@ typedef struct IMG_Animation
 } IMG_Animation;
 
 /**
+ * Detect if a WEBP image is animated.
+ *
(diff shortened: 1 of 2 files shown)


slouken:
What's the use case for this?

LordOfXen:
> What's the use case for this?

Simplifies API for checking WebP format for animation since SDL has isPNG etc., detecting between animated or not is a great addition in case user has a checkwebp function, he wouldn't have to load the entire WebP image.

This also allows early checks where the image wouldn't be loaded yet, so the code knows where to forward the image data (stream or whatever they hold) to for loading.

LordOfXen:
Closing for consistency with the rest of the API, feel free to re-open for a global/format-independent solution.

PR #577, opened 2025-08-01
Added an API for saving a multi-frame animation

slouken:


slouken:
@icculus, @sezero, @LordOfXen for review

LordOfXen:
Looks quite good, however there are linker errors I think, I will take a look in a little bit.
Edit: Errors are unrelated to this pull request I believe.

src/IMG_anim.c
```diff
@@ -0,0 +1,195 @@
...
+
+    if (!type) {
+        if (file) {
+            type = SDL_strrchr(file, '.');
+            if (type) {
+                // Skip the '.' in the file extension
+                ++type;
+            }
+        }
+        if (!type) {
```
sezero: I guess `!type || !*type` would be safer

examples/showanim.c
```diff
@@ -98,6 +119,12 @@ int main(int argc, char *argv[])
             continue;
         }
 
+        if (SDL_strcmp(argv[i], "-save") == 0 && argv[i + 1]) {
+            save = argv[i + 1];
+            ++i;
```
sezero, line 123: For sake of nitpicking: `save = argv[++i];` maybe ?

include/SDL3_image/SDL_image.h
```diff
@@ -2179,6 +2179,111 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+ * - `IMG_PROP_ANIMATION_STREAM_CREATE_IOSTREAM_POINTER`: an SDL_IOStream that will be used to save the stream. This should not be closed until the animation stream is closed. This is required if `IMG_PROP_ANIMATION_STREAM_CREATE_FILENAME_STRING` isn't set.
+ * - `IMG_PROP_ANIMATION_STREAM_CREATE_IOSTREAM_AUTOCLOSE_BOOLEAN`: true if closing the animation stream should also close the associated SDL_IOStream.
+ * - `IMG_PROP_ANIMATION_STREAM_CREATE_TYPE_STRING`: the output file type, e.g. "webp", defaults to the file extension if `IMG_PROP_ANIMATION_STREAM_CREATE_FILENAME_STRING` is set.
+ * - `IMG_PROP_ANIMATION_STREAM_CREATE_QUALITY_NUMBER`: the compression quality, in the range of 0 to 100. The higher the number, the higher the quality and file size. This defaults to a balanced value for compression and quality.
+ * - `IMG_PROP_ANIMATION_STREAM_CREATE_TIMEBASE_NUMERATOR_NUMBER`: the numerator of the fraction used to multiply the pts to convert it to seconds. This defaults to 1.
+ * - `IMG_PROP_ANIMATION_STREAM_CREATE_TIMEBASE_DENOMINATOR_NUMBER`: the denominator of the fraction used to multiply the pts to convert it to seconds. This defaults to 1000.
```
sezero, line 2235: I hope wiki shit doesn't screw up this list (CC @icculus )

sezero:
Xcode and cmake sides need updting for the new libwebpmux dependency (CC @madebr )

src/IMG_anim.c
```diff
@@ -0,0 +1,195 @@
...
+    IMG_AnimationStream *stream = (IMG_AnimationStream *)SDL_calloc(1, sizeof(*stream));
+    if (!stream) {
+        goto error;
+    }
+
+    stream->dst = dst;
+    stream->start = SDL_TellIO(dst);
+    stream->closeio = closeio;
+    stream->quality = (int)SDL_GetNumberProperty(props, IMG_PROP_ANIMATION_STREAM_CREATE_QUALITY_NUMBER, -1);
+    stream->timebase_numerator = timebase_numerator;
```
LordOfXen, line 131: Think `timebase_numerator` and `timebase_denominator` helper would be good, since some formats only takes int, we will have to spread the calculation around for each format that asks for integer, instead we can have an internal function to get integer maybe that reduces dependency on those variables, in which case if we modify this later, we wouldn't touch the code in formats.

examples/showanim.c
slouken: The existing code is a little more clear.

include/SDL3_image/SDL_image.h
slouken: It shouldn't.

src/IMG_anim.c
slouken: I'm not sure what you mean?

LordOfXen: Some formats like gif asks for integer, not numerator and denominator, which means we will have to do a conversion to an integer value containing milliseconds, to reduce code complexity maybe moving that into IMG_anim.h would be better since that's not only for gif itself, we would be coding the same for similar formats as well.

LordOfXen: Something like this:
```c
  GetIntegerMS()
  {
      int retval;
      if (denominator == 0) {
          retval = numerator * 100;
      } else {
          retval = (int)(((float)numerator / denominator) * 1000.0f);
      }
      return retval;
  }
```

slouken: Oh, that's not actually how the PTS stuff works. Take a look at the webp usage of those fields.

slouken:
@madebr, I took a stab at adding webpmux to the cmake support, but I've bunged it up. Can you take a look?

sezero:
Note: Is libwebpmux always enabled in recent distros? In old libwebp-1.0.3, for e.g., I had to pass `--enable-libwebpmux` to configure. (Haven't looked at cmake. Haven't looked at more recent libwebp releases.)

src/IMG_anim.c
LordOfXen: I checked your code for the new functions for pushing frames from WEBP to encode animation, and it does similar to what I posted above.

`  int timestamp = (int)((pts * 1000 * stream->timebase_numerator) / stream->timebase_denominator);`

I mean that's fine to do, but this will be done for all formats requiring int is what I tried to say, so we will do this for them all over and over. However, it's a slight improvement to standardize it so it's not mandatory, depends on your thoughts.

CMakeLists.txt
```diff
@@ -978,17 +979,19 @@ if(SDLIMAGE_WEBP)
         target_include_directories(${sdl3_image_target_name} PRIVATE external/libwebp/src)
         add_library(WebP::webp ALIAS webp)
         add_library(WebP::webpdemux ALIAS webpdemux)
+        add_library(WebP::webpmux ALIAS webpmux)
```
madebr: I think it has a [`lib` prefix](https://github.com/libsdl-org/libwebp/blob/fa992316e4517b25a4e667a3668b930170becd56/CMakeLists.txt#L528C3-L528C14).
```suggestion
        add_library(WebP::webpmux ALIAS libwebpmux)
```

CMakeLists.txt
```diff
@@ -978,17 +979,19 @@ if(SDLIMAGE_WEBP)
...
         if(SDLIMAGE_WEBP_SHARED OR NOT SDLIMAGE_BUILD_SHARED_LIBS)
-            list(APPEND INSTALL_EXTRA_TARGETS webp webpdemux)
+            list(APPEND INSTALL_EXTRA_TARGETS webp webpdemux webpmux)
```
madebr:
```suggestion
            list(APPEND INSTALL_EXTRA_TARGETS webp webpdemux libwebpmux)
```

CMakeLists.txt
```diff
@@ -978,17 +979,19 @@ if(SDLIMAGE_WEBP)
...
             if(NOT SDLIMAGE_WEBP_SHARED)
                 list(APPEND INSTALL_EXTRA_TARGETS sharpyuv)
             endif()
         endif()
         set_target_properties(webp PROPERTIES EXPORT_NAME "external_libwebp")
         set_target_properties(webpdemux PROPERTIES EXPORT_NAME "external_webpdemux")
+        set_target_properties(webpmux PROPERTIES EXPORT_NAME "external_webpmux")
```
madebr:
```suggestion
        set_target_properties(libwebpmux PROPERTIES EXPORT_NAME "external_libwebpmux")
```

LordOfXen:
> Note: Is libwebpmux always enabled in recent distros? In old libwebp-1.0.3, for e.g., I had to pass `--enable-libwebpmux` to configure. (Haven't looked at cmake. Haven't looked at more recent libwebp releases.)

I tried to take a look with Windows subsystem for linux (Ubuntu 24.04.2) and installed libwebp-dev, it downloaded the followings:
Get:1 http://archive.ubuntu.com/ubuntu noble/main amd64 libsharpyuv-dev amd64 1.3.2-0.4build3 [16.0 kB]
Get:2 http://archive.ubuntu.com/ubuntu noble/main amd64 libwebpdecoder3 amd64 1.3.2-0.4build3 [114 kB]
Get:3 http://archive.ubuntu.com/ubuntu noble/main amd64 libwebpdemux2 amd64 1.3.2-0.4build3 [12.4 kB]
Get:4 http://archive.ubuntu.com/ubuntu noble/main amd64 libwebpmux3 amd64 1.3.2-0.4build3 [25.7 kB]
Get:5 http://archive.ubuntu.com/ubuntu noble/main amd64 libwebp-dev amd64 1.3.2-0.4build3 [367 kB]

Seems to gather mux3 directly.

CMakeLists.txt
sezero: > I think it has a `lib` prefix

Its brothers above don't ??

madebr: > Its brothers above don't ??

Tell that to the Google devs ;)
The source never lies!!

slouken:
@madebr, those changes don't fix anything, I'm still getting this error:
```
CMake Error at cmake/PrivateSdlFunctions.cmake:103 (get_target_property):
  get_target_property() called with non-existent target "WebP::webpmux".
  cmake/PrivateSdlFunctions.cmake:121 (get_actual_target)
  CMakeLists.txt:1032 (target_get_dynamic_library)
```

CMakeLists.txt
madebr: @slouken
There are multiple occurrences of `webpmux`  that need to renamed to `libwebpmux`.

slouken: Ah, got it, thanks!

slouken: Can you check and make sure I changed the right ones?

sezero:
> > Note: Is libwebpmux always enabled in recent distros? In old libwebp-1.0.3, for e.g., I had to pass `--enable-libwebpmux` to configure. (Haven't looked at cmake. Haven't looked at more recent libwebp releases.)
> 
> I tried to take a look with Windows subsystem for linux (Ubuntu 24.04.2) and installed libwebp-dev, it downloaded the followings:
> [...]
> Seems to gather mux3 directly.

OK, rpm search shows it is available in fedora and the likes:
https://rpmfind.net/linux/rpm2html/search.php?query=libwebpmux.so&submit=Search+...

Even in EOL distros like rhel / centos 6 and 7, it is available.Therefore things should be OK.

src/IMG_anim.c
slouken: Yep, good idea. Done!

sezero:
I had to do the following, otherwise it failed to find the system-installed libwebpmux:

```diff
diff --git a/cmake/Findwebp.cmake b/cmake/Findwebp.cmake
index b719fb2..ce1a1be 100644
--- a/cmake/Findwebp.cmake
+++ b/cmake/Findwebp.cmake
@@ -28,22 +28,22 @@ set(webpdemux_LINK_LIBRARIES "" CACHE STRING "Extra link libraries of webpdemux"
 
 set(webpdemux_LINK_FLAGS "" CACHE STRING "Extra link flags of webpdemux")
 
-find_library(libwebpmux_LIBRARY
-    NAMES libwebpmux
+find_library(webpmux_LIBRARY
+    NAMES webpmux
 )
 
-find_path(libwebpmux_INCLUDE_PATH
+find_path(webpmux_INCLUDE_PATH
     NAMES webp/mux.h
 )
 
-set(libwebpmux_COMPILE_OPTIONS "" CACHE STRING "Extra compile options of webpmux")
+set(webpmux_COMPILE_OPTIONS "" CACHE STRING "Extra compile options of webpmux")
 
-set(libwebpmux_LINK_LIBRARIES "" CACHE STRING "Extra link libraries of webpmux")
+set(webpmux_LINK_LIBRARIES "" CACHE STRING "Extra link libraries of webpmux")
 
-set(libwebpmux_LINK_FLAGS "" CACHE STRING "Extra link flags of webpmux")
+set(webpmux_LINK_FLAGS "" CACHE STRING "Extra link flags of webpmux")
 
 find_package_handle_standard_args(webp
-    REQUIRED_VARS webp_LIBRARY webp_INCLUDE_PATH webpdemux_LIBRARY webpdemux_INCLUDE_PATH libwebpmux_LIBRARY libwebpmux_INCLUDE_PATH
+    REQUIRED_VARS webp_LIBRARY webp_INCLUDE_PATH webpdemux_LIBRARY webpdemux_INCLUDE_PATH webpmux_LIBRARY webpmux_INCLUDE_PATH
 )
 
 if (webp_FOUND)
@@ -70,11 +70,11 @@ if (webp_FOUND)
     if (NOT TARGET WebP::libwebpmux)
         add_library(WebP::libwebpmux UNKNOWN IMPORTED)
         set_target_properties(WebP::libwebpmux PROPERTIES
-            IMPORTED_LOCATION "${libwebpmux_LIBRARY}"
-            INTERFACE_INCLUDE_DIRECTORIES "${libwebpmux_INCLUDE_PATH}"
-            INTERFACE_COMPILE_OPTIONS "${libwebpmux_COMPILE_FLAGS}"
-            INTERFACE_LINK_LIBRARIES "${libwebpmux_LINK_LIBRARIES}"
-            INTERFACE_LINK_FLAGS "${libwebpmux_LINK_FLAGS}"
+            IMPORTED_LOCATION "${webpmux_LIBRARY}"
+            INTERFACE_INCLUDE_DIRECTORIES "${webpmux_INCLUDE_PATH}"
+            INTERFACE_COMPILE_OPTIONS "${webpmux_COMPILE_FLAGS}"
+            INTERFACE_LINK_LIBRARIES "${webpmux_LINK_LIBRARIES}"
+            INTERFACE_LINK_FLAGS "${webpmux_LINK_FLAGS}"
         )
     endif()
 endif()
```

cmake/Findwebp.cmake
```diff
@@ -28,8 +28,22 @@ set(webpdemux_LINK_LIBRARIES "" CACHE STRING "Extra link libraries of webpdemux"
 
 set(webpdemux_LINK_FLAGS "" CACHE STRING "Extra link flags of webpdemux")
 
+find_library(libwebpmux_LIBRARY
+    NAMES libwebpmux
```
madebr: Remove the `lib` prefix: this would be looking for `liblibwebpmux.so` otherwise.
```suggestion
    NAMES webpmux
```

madebr:
Will there be an `SDLIMAGE_WEBP_SAVE` option, similarly to `SDLIMAGE_(AVIF|JPG|PNG)_SAVE`?

slouken:
> Will there be an `SDLIMAGE_WEBP_SAVE` option, similarly to `SDLIMAGE_(AVIF|JPG|PNG)_SAVE`?

That's probably a good idea, but should come after this commit, since it will make webpmux optional.

sezero:
Fix for `SDLIMAGE_DEPS_SHARED=OFF` configuration:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 904393f..38e1a97 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1036,7 +1036,7 @@ if(SDLIMAGE_WEBP)
             message(STATUS "Dynamic libwebp: ${SDLIMAGE_DYNAMIC_WEBP}")
             target_compile_definitions(${sdl3_image_target_name} PRIVATE "LOAD_WEBP_DYNAMIC=\"${SDLIMAGE_DYNAMIC_WEBP}\"")
         else()
-            target_link_libraries(${sdl3_image_target_name} PRIVATE WebP::webp WebP::webpdemux)
+            target_link_libraries(${sdl3_image_target_name} PRIVATE WebP::webp WebP::webpdemux WebP::libwebpmux)
         endif()
     endif()
 endif()
```

LordOfXen:
Shouldn't we have width and height in IMG_AnimationStream as well? All animations should have that anyway for the desired width and height, as well as a desired pixel format.

DesiredWidth: -1 or actual value
DesiredHeight: -1 or actual value
DesiredPixelFormat: -1 or actual value

If the encoder receives -1 for width but the actual value for height (or vice versa), it can resize the surfaces with the given axis. If nothing is defined or one is undefined as specified in the sentence, we get the width or height or both from the first frame/first call to AddFrame.

If the encoder receives the desired pixel format of -1, then it can first check if the pixel format of the first frame is valid for the underlying encoder, then decide to use it or assign a default pixel format in cases where the first frame's pixel format is not valid (for the underlying encoder).

LordOfXen:
Also what is the unit for pst? For formats where they directly ask for numerator and denominator for each frame, we need to get that from Uint64 pst that's given to AddFrame function of each format. And for that we need to know the frequency of pts.

Something like this needed probably
`#define PTS_TIMEBASE_UNITS_PER_SECOND 1000000ULL`

src/IMG_anim.c
```diff
@@ -0,0 +1,200 @@
...
+    }
+    pts -= stream->first_pts;
+
+    // Make sure the presentation timestamp isn't out of order
+    if (pts < stream->last_pts) {
+        return SDL_SetError("Frame added out of order");
+    }
+    stream->last_pts = pts;
+
+    return stream->AddFrame(stream, surface, pts);
```
LordOfXen: I have noticed a logical flaw. stream->last_pts must be set **after** the call to AddFrame, otherwise, you can't get access to the previous pts. If it stays like this, then it carries unnecessary data as the AddFrame already holds the current/last pts as its last parameter.

LordOfXen:
Another addition would be another overload for IMG_CreateAnimationStream and IMG_CreateAnimationStream_IO that takes SDL_PropertiesID for extra properties of encoders in order to reduce code complexity on user side.

Briefly for this usage:

 ```c
SDL_PropertiesID customProps = SDL_CreateProperties();
SDL_SetNumberProperty(customProps, "webp_numplays", 5); // e.g. user just wants to set how many times the generated webp should play but nothing else.

IMG_AnimationStream* animStream = IMG_CreateAnimationStream("myimage.webp", customProps); // Easy construction without the hassle with default values that SDL already handles.
```

slouken:
> Shouldn't we have width and height in IMG_AnimationStream as well? All animations should have that anyway for the desired width and height, as well as a desired pixel format.
> 
> DesiredWidth: -1 or actual value DesiredHeight: -1 or actual value DesiredPixelFormat: -1 or actual value

No, these are inferred from the frames being fed into the encoder. If the application wants a specific size or format, it can convert the frames before feeding them into the stream.

src/IMG_anim.c
slouken: Sure, that's more clear that last_pts is always the last presentation timestamp. Fixed.

slouken:
> Another addition would be another overload for IMG_CreateAnimationStream and IMG_CreateAnimationStream_IO that takes SDL_PropertiesID for extra properties of encoders in order to reduce code complexity on user side.

No, if you start using properties, you just use IMG_CreateAnimationStreamWithProperties(). Most of the time the application won't bother with that level of complexity, or if they do, they will wrap it in a simple function they implement.

LordOfXen:
> > Shouldn't we have width and height in IMG_AnimationStream as well? All animations should have that anyway for the desired width and height, as well as a desired pixel format.
> > DesiredWidth: -1 or actual value DesiredHeight: -1 or actual value DesiredPixelFormat: -1 or actual value
> 
> No, these are inferred from the frames being fed into the encoder. If the application wants a specific size or format, it can convert the frames before feeding them into the stream.

Then the encoder will fail if any of the frames doesn't match the first given frame's width/height/format, just clarifying because this changes the path towards newer encoder codes.

slouken:
> > > Shouldn't we have width and height in IMG_AnimationStream as well? All animations should have that anyway for the desired width and height, as well as a desired pixel format.
> > > DesiredWidth: -1 or actual value DesiredHeight: -1 or actual value DesiredPixelFormat: -1 or actual value
> > 
> > 
> > No, these are inferred from the frames being fed into the encoder. If the application wants a specific size or format, it can convert the frames before feeding them into the stream.
> 
> Then the encoder will fail if any of the frames doesn't match the first given frame's width/height/format, just clarifying because this changes the path towards newer encoder codes.

Yes, that's fine. We can either document that the frames must match sizes, or resize frames to match the first one. I'm leaning towards just documenting that, at least for now, and we can change that later.

slouken:
Thanks for the help everyone!

PR #584, opened 2025-08-04
Destroy image if NormalizeFrames fails in GIF Animation

### LINKED ISSUE #583: Potential Memory Leak in NormalizeFrames for GIF
I believe I've noticed a memory leak in the NormalizeFrames function.
Please check below:

### Create image
Image gets created here:
```c
if (SDL_SurfaceHasColorKey(frames[0].image)) {
    image = SDL_ConvertSurface(frames[0].image, SDL_PIXELFORMAT_ARGB8888); // either here
} else {
    image = SDL_ConvertSurface(frames[0].image, SDL_PIXELFORMAT_XRGB8888); // or here
}
```

### Try to duplicate surface
Here it tries to duplicate the surface; if it fails to do so, it returns directly without a call to SDL_DestroySurface for the actual 'image' variable that was allocated/created above.
```c
frames[i].image = SDL_DuplicateSurface(image);
if (!frames[i].image) {
    return false;   // <-- Returns directly here without destroying the image.
}
```

SDL_DuplicateSurface isn't going to fail most of the time, but if it does, we are going to leak memory silently.

LordOfXen:
Destroys the created image surface if NormalizeFrames fails so we don't silently leak memory.

This fixes #583

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -188,6 +188,7 @@ static bool NormalizeFrames(Frame_t *frames, int count)
         SDL_DestroySurface(frames[i].image);
         frames[i].image = SDL_DuplicateSurface(image);
         if (!frames[i].image) {
+            SDL_DestroySurface( image );
             return false;
         }
 


slouken:
Merged, thanks!

LordOfXen:
> Merged, thanks!

A pleasure!

PR #585, opened 2025-08-04
Support for Encoding GIFs

LordOfXen:
Support for encoding `SDL_Surface` as `GIF` to an `SDL_IOStream` with the [Animation Stream](https://github.com/libsdl-org/SDL_image/pull/577) feature of `SDL_image`.
**_This addition does not touch the existing `GIF` loader code and only adds the new encoding feature on top of it._**

### Current features
- Efficient compression using the custom `lzwCompress` function.
- `SAVE_GIF_OCTREE` flag that can be set to 0 to use SDL's indexing for non-indexed surfaces. By default, for quality, it uses octree quantization.
- Follow SDL's previous implementation for loading GIFs if `LOAD_GIF` is declared, a similar flag `SAVE_GIF` has been introduced.
- Allows setting transparent_color_index through the passed `SDL_PropertiesID`.
- Allows setting num_colors (between 2 and up to 256 by power of two) through the passed `SDL_PropertiesID`.
- Allows setting loop_count through the passed `SDL_PropertiesID`.

### What to consider?
- Currently, it allocates the buffer for the bit stream twice the current size if it's small for the next bit stream code. This might be problematic with large images, since it will allocate more memory than needed.  Although it is not a big concern, we could implement a more sophisticated resizing strategy.
- Available settings must be mapped into defines and moved into SDL_image.h so it is accessible directly; this has to be discussed.

### Developer notes
Currently, this does not introduce any dithering (e.g., Jarvis/FloydSteinberg dithering) because it allows SDL_PIXELFORMAT_INDEX8 to be passed directly and mapped to the output indexed pixels buffer. This means users can modify the buffer with SDL before passing it to the encoder, eliminating the need for mandatory dithering within the encoder and also reducing complexity in the source code.

**_Please note that the octree quantization works only if SAVE_GIF_OCTREE is set to 1 and if the processed surfaces are not indexed_**.

### Available Tests
Tested in Windows 11 24H2 and seems to work as intended.

### Original Image:
test

### Decode the Original GIF and Encode Back
save_gif_test

### Decode the Encoded GIF and Encode Again to Confirm Encoding is Correct
save_gif_test_validate

### Binary Validation - Comparison
The original binary matches the binary encoded by our GIF encoder, confirming the writing is correct.

### Original GIF Binary:
```
GIF size: 1,520,180 bytes (1.45mb)
GIF length: 6.5 seconds
GIF width/height: 800×600 pixels
number of frames: 325
number of colors: 256
loop count: 0 (endless)
frames 1-325: 20ms

First Frame:
-------------
x: 0
y: 0
width: 800
height: 600
delay: 20ms
disposal: 1
transparent: false
transparency index: 0 (not used)
transparency color: no (not used)
```

### Encoded GIF Binary
```
GIF size: 1,209,183 bytes (1.15mb)
GIF length: 6.5 seconds
GIF width/height: 800×600 pixels
number of frames: 325
number of colors: 256
loop count: 0 (endless)
frames 1-325: 20ms

First Frame:
-------------
x: 0
y: 0
width: 800
height: 600
delay: 20ms
disposal: 1
transparent: false
transparency index: 0 (not used)
transparency color: no (not used)
```

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -2223,11 +2223,6 @@ static int writeGifTrailer(SDL_IOStream *io)
 
 static bool AnimationStream_AddFrame(struct IMG_AnimationStream *stream, SDL_Surface *surface, Uint64 pts)
 {
-    if (!stream || !stream->ctx || !surface) {
-        SDL_SetError("Invalid arguments for AnimationStream_AddFrame.");
-        return false;
-    }
-
     IMG_AnimationStreamContext *ctx = stream->ctx;
     SDL_IOStream *io = stream->dst;
     uint8_t *indexedPixels = NULL;


sezero:
Here is a patch that removes imcompatble pointer warnings (removal of those consts), fixing type redefinition errors, and ensuring gif stucture packing.

```diff
diff --git a/src/IMG_gif.c b/src/IMG_gif.c
index 7c90f61..386aeda 100644
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -822,6 +822,7 @@ SDL_Surface *IMG_LoadGIF_IO(SDL_IOStream *src)
 #if defined(SAVE_GIF)
 // GIF Header (6 bytes) + Logical Screen Descriptor (7 bytes)
 // Total 13 bytes
+#pragma pack(push,1)
 typedef struct
 {
     char signature[3]; // "GIF"
@@ -869,6 +870,7 @@ typedef struct
     uint16_t loopCount;                    // 0 for infinite loop
     uint8_t blockTerminator;               // 0x00
 } NetscapeExtension;
+#pragma pack(pop)
 
 
 // Function to write a byte to the SDL_IOStream
@@ -886,7 +888,7 @@ static bool writeWord(SDL_IOStream *io, uint16_t word)
     return SDL_WriteIO(io, bytes, 2) == 2;
 }
 
-typedef struct IMG_AnimationStreamContext
+struct IMG_AnimationStreamContext
 {
     uint16_t width;
     uint16_t height;
@@ -897,7 +899,7 @@ typedef struct IMG_AnimationStreamContext
     bool firstFrame;
     uint8_t colorMapLUT[32][32][32];
     bool lut_initialized;
-} IMG_AnimationStreamContext;
+};
 
 #define LZW_MAX_CODES 4096
 #define LZW_MAX_BITS  12
@@ -1494,7 +1496,7 @@ static int count_set_bits(uint32_t n)
     return count;
 }
 
-static void buildColorMapLUT(uint8_t lut[32][32][32], const uint8_t palette[][3], uint16_t numColors, bool hasTransparency)
+static void buildColorMapLUT(uint8_t lut[32][32][32], uint8_t palette[][3], uint16_t numColors, bool hasTransparency)
 {
     const int color_count = hasTransparency ? numColors - 1 : numColors;
 
@@ -1528,7 +1530,7 @@ static void buildColorMapLUT(uint8_t lut[32][32][32], const uint8_t palette[][3]
     }
 }
 
-static int mapSurfaceToExistingPalette(SDL_Surface *psurf, const uint8_t lut[32][32][32], uint8_t *indexedPixels, int transparentIndex)
+static int mapSurfaceToExistingPalette(SDL_Surface *psurf, uint8_t lut[32][32][32], uint8_t *indexedPixels, int transparentIndex)
 {
     SDL_Surface *surf = psurf;
     bool surface_converted = false;
@@ -2015,12 +2017,12 @@ static int writeGifHeader(SDL_IOStream *io, uint16_t width, uint16_t height,
     return 0;
 }
 
-static int writeColorTable(SDL_IOStream *io, const uint8_t colors[][3], uint16_t numColors)
+static int writeColorTable(SDL_IOStream *io, uint8_t colors[][3], uint16_t numColors)
 {
     if (!io || !colors || numColors == 0)
         return -1;
 
-    if (SDL_WriteIO(io, colors, numColors * 3) != numColors * 3) {
+    if (SDL_WriteIO(io, colors, numColors * 3) != numColors * 3u) {
         SDL_SetError("Failed to write color table data");
         return -1;
     }
```

LordOfXen:
> Here is a patch that removes imcompatble pointer warnings (removal of those consts), fixing type redefinition errors, and ensuring gif stucture packing.

Thanks! Fixed now.

src/IMG_gif.c
```diff
@@ -857,6 +857,7 @@ typedef struct
 } ImageDescriptor;
 
 // Netscape Application Extension (19 bytes + 1 byte terminator)
+#pragma pack(push,1)
```
sezero: This can (should) be moved above to line 823 in order to ensure all those structs' packing / sizes.

LordOfXen: Sure, I will cover up to GifHeaderAndLSD.

sezero:
I _think_ this should fix builds targeting Apple OSs:

```diff
diff --git a/src/IMG_gif.c b/src/IMG_gif.c
index 1705c82..ff2b922 100644
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -2433,4 +2433,13 @@ bool IMG_CreateGIFAnimationStream(struct IMG_AnimationStream *stream, SDL_Proper
 #endif /*!defined(SAVE_GIF)*/
 }
 
+#else
+
+bool IMG_CreateGIFAnimationStream(struct IMG_AnimationStream *stream, SDL_PropertiesID props)
+{
+    (void)stream;
+    (void)props;
+    return SDL_SetError("GIF animation saving is not enabled in this build.");
+}
+
 #endif /* !defined(__APPLE__) || defined(SDL_IMAGE_USE_COMMON_BACKEND) */
```

sezero:
> I _think_ this should fix builds targeting Apple OSs:

Note though, if gif saving ever gets implemented in IMG_imageio, then the problem will disappear.

LordOfXen:
@sezero Actually, wouldn't it be better to keep the new saving code _outside_ the scope of `!defined(__APPLE__) || defined(SDL_IMAGE_USE_COMMON_BACKEND)`?

That scope only contains IMG_LoadGIF_IO and IMG_isGIF, not the actual internal functions as far as I can see.

LordOfXen:
> Note though, if gif saving ever gets implemented in IMG_imageio, then the problem will disappear.

That is probably the fix yes, but please consider my previous question.

sezero:
> @sezero Actually, wouldn't it be better to keep the code _outside_ the scope of `!defined(__APPLE__) || defined(SDL_IMAGE_USE_COMMON_BACKEND)`?
> 
> That scope only contains IMG_LoadGIF_IO and IMG_isGIF, not the actual internal functions as far as I can see.

That works too, I think.

(I wish IMG_imageio and IMG_wic would someday be removed, though..)

LordOfXen:
> (I wish IMG_imageio and IMG_wic would someday be removed, though..)

That'd be better, but it's sadly outside the scope of this PR and maybe we can make another PR later regarding this.

src/IMG_gif.c
```diff
@@ -809,3 +820,1617 @@ SDL_Surface *IMG_LoadGIF_IO(SDL_IOStream *src)
...
+    if (!stream) {
+        SDL_SetError("Invalid IMG_AnimationStream pointer.");
+        return false;
+    }
+
+    if (!stream->dst) {
+        SDL_SetError("SDL_IOStream (stream->dst) is NULL. It must be provided by the caller.");
+        return false;
+    }
+
```
slouken:
```suggestion
```

These are checked by calling code and do not need to be double checked.

src/IMG_gif.c
```diff
@@ -809,3 +820,1607 @@ SDL_Surface *IMG_LoadGIF_IO(SDL_IOStream *src)
...
+    return 0;
+}
+
+static bool AnimationStream_AddFrame(struct IMG_AnimationStream *stream, SDL_Surface *surface, Uint64 pts)
+{
+    if (!stream || !stream->ctx || !surface) {
+        SDL_SetError("Invalid arguments for AnimationStream_AddFrame.");
+        return false;
+    }
+
```
slouken:
```suggestion
```

Also all guaranteed to be valid...

slouken:
Merged, thanks!

slouken:
Now that we have GIF saving code, does it make sense to add a function to save a single GIF image?

LordOfXen:
> Now that we have GIF saving code, does it make sense to add a function to save a single GIF image?

It's actually not needed in my opinion, user can only push a single frame then close the stream if he wants to encode a single frame.

Or we can make a single function in IMG_anim.c that wraps all those open push close functions for single frame encoding (note that it would be an unnecessary step for formats like APNG, so it is better for user to do this for specific formats like GIF).

slouken:
> Or we can make a single function in IMG_anim.c that wraps all those open push close functions for single frame encoding (note that it would be an unnecessary step for formats like APNG, so it is better for user to do this for specific formats like GIF).

Yes, let's do that, for symmetry with other format save functions.

slouken:
(but the function should go in IMG_gif.c)

LordOfXen:
> (but the function should go in IMG_gif.c)

Sure that would work, I will add that in a bit.

PR #587, opened 2025-08-05
[SDL2] Support library name variation in pkg-config

FtZPetruska:
When building with CMake, the library may have its final name altered in a variety of ways:
- a debug postfix `d`, added by default (but which can be easily removed by setting `SDL2IMAGE_DEBUG_POSTFIX`),
- on OS/2, the shared library is named `SDL2img`,
- on MSVC and Watcom, the static library gets a `-static` suffix.

To account for these variation, I made library name in the pkg-config module configurable (through `PC_LIBNAME`), for CMake builds it's set to the result of the `TARGET_LINKER_FILE_BASE_NAME` generator expression which covers changes to `OUTPUT_NAME` and `<CONFIG>_POSTFIX`.

A notable change is that on platforms where the output name of the shared library is `SDL2_image-${LT_RELEASE}`, that is the name that gets picked up. In practice I did not find it to be a meaningful difference, linking to `-lSDL2_image` or `-lSDL2_image-2.0` will both lead to the exact same executable being produced.

For context, the main issue I was personally facing is the `-static` suffix in MSVC static builds.  

---

NB: I'm not used to working with autotools, so apologies if this is not how it's supposed to be regenerated.


sezero:
This I don't like at all: corner cases that only a few would care, if at all. I do suggest leaving things as they are.

sezero (commented):
This I don't like at all: corner cases that only a few would care, if at all. I do suggest leaving things as they are.

madebr:
> This I don't like at all: corner cases that only a few would care, if at all. I do suggest leaving things as they are.

I tend to agree. I consider this `d` suffix of the SDL2 libraries a documented defect, fixed in SDL3

sezero:
OK, closing.

FtZPetruska:
Thank you all for the quick feedback!

PR #589, opened 2025-08-05
Enhance TGA Saving Functions

LordOfXen:
Added SAVE_TGA for TGA saving functions so they are not merged in LOAD_TGA.


sezero:
I _think_ you should move SAVE_TGA functionality out of `__APPLE__` defines, because such funtionality isn't implemented in IMG_ImageIO.m

CC @slouken

sezero:
NOTE: Unlike others, SaveTGA_IO doesn't have a `closeio` parameter, _and_ it doesn't document whether or not it keeps the iostream open after returning.

LordOfXen:
@sezero latest commits fixes all of the ones you mentioned. Please take a look.

slouken:
Merged, thanks!

PR #591, opened 2025-08-05
Encode & Decode AVIF Animation Support (AVIFS)

LordOfXen:
## Introducing AVIF Animation Decoding and Encoding Support (aka AVIFS)
Implements `IMG_CreateAVIFAnimationStream` for encoding and `IMG_LoadAVIFAnimation_IO` for decoding `AVIFS` files similar to `WEBP`, `GIF`, and `APNG` support.

_**Please note that the AVIFS encoding/decoding does not modify the existing AVIF code and only adds this new support on top of it.**_

### Current Features
- Exposes `maxthreads` through the supplied `SDL_PropertiesID` so users can pick between 1 and the max available logical CPU cores for encoding.
- Exposes `keyframeinterval` through the supplied `SDL_PropertiesID` so users can decide between 1 and the desired value for lossless, and between 0 and the desired value for lossy encodings. Please note that for lossy encoding, it currently does not set the keyframe and only enforces a keyframe interval to be 1 or above for lossless encoding. When lossy is picked and the keyframe is set to 0, you may see dirty region artifacts; this is usually due to AVIF's internal encoding. To prevent this, use lossless encoding where possible, or just keyframe to 1.
- Support for 16-bit images (`SDL_PIXELFORMAT_RGB48`, `SDL_PIXELFORMAT_BGR48`, `SDL_PIXELFORMAT_RGBA64`, `SDL_PIXELFORMAT_ARGB64`, `SDL_PIXELFORMAT_BGRA64`, and `SDL_PIXELFORMAT_ABGR64`) in both encoder and decoder.
- Support for 10-bit images (`SDL_ISPIXELFORMAT_10BIT`) in both encoder and decoder.
- Support for HDR with the `SetHDRProperties` helper function in both encoder and decoder. Respects `SDL_PROP_SURFACE_MAXCLL_NUMBER`, `SDL_PROP_SURFACE_MAXFALL_NUMBER`, `SDL_PROP_SURFACE_SDR_WHITE_POINT_FLOAT`, and `SDL_PROP_SURFACE_HDR_HEADROOM_FLOAT`.
- Support for SDR (8-bit) images (`RGB`, `BGR`, `RGBA`, and `BGRA`) in both encoder and decoder.
- Ignores the alpha channel if not present for efficiency in encoding/decoding.

This has been achieved using the existing libavif16 and the pre-existing encode/decode functions in the `lib` struct; nothing additional has been introduced.

### DIFF
--- a/src/IMG_avif.c
+++ b/src/IMG_avif.c
@@ -22,6 +22,7 @@
 /* This is a AVIF image file loading framework */
 
 #include <SDL3_image/SDL_image.h>
+#include "IMG_anim.h"
 
 /* We'll have AVIF save support by default */
 #if !defined(SDL_IMAGE_SAVE_AVIF)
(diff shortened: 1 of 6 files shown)


LordOfXen:
@slouken @sezero for review.

src/IMG_avif.c
```diff
@@ -756,3 +757,785 @@ bool IMG_SaveAVIF_IO(SDL_Surface *surface, SDL_IOStream *dst, bool closeio, int
...
+            }
+
+            if (!frame_surface) {
+                SDL_SetError("Couldn't create 16-bit surface for AVIF frame");
+                error = true;
+                goto done;
+            }
+
+            bool lockedSurf = false;
+            if (SDL_MUSTLOCK(frame_surface)) {
```
slouken: This will never be true for a newly created surface.

src/IMG_avif.c
```diff
@@ -756,3 +757,785 @@ bool IMG_SaveAVIF_IO(SDL_Surface *surface, SDL_IOStream *dst, bool closeio, int
...
+                SDL_SetError("Couldn't convert 16-bit AVIF image to RGB: %s", lib.avifResultToString(result));
+                SDL_DestroySurface(frame_surface);
+                frame_surface = NULL;
+                error = true;
+                goto done;
+            }
+
+             SetHDRProperties(frame_surface, image);
+        }
+        else if (image->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084) {
```
slouken:
```suggestion
        } else if (image->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084) {
```

src/IMG_avif.c
```diff
@@ -756,3 +757,785 @@ bool IMG_SaveAVIF_IO(SDL_Surface *surface, SDL_IOStream *dst, bool closeio, int
...
+IMG_Animation* IMG_LoadAVIFAnimation_IO(SDL_IOStream* src)
+{
+    (void)src;
+    SDL_SetError("SDL_image built without AVIF animation loading support");
+    return NULL;
+}
+
+#endif /* LOAD_AVIF */
+
+#if SDL_IMAGE_SAVE_AVIF
```
slouken, line 1058:
```suggestion
#if SAVE_AVIF
```
The other formats don't use the `SDL_IMAGE_` prefix. We could add that as a separate commit if we really wanted.

LordOfXen:
@slouken Just for clarity, SDL_ConvertSurface returns a new surface entirely right? Since a newly created surface will not be evaluated to true within SDL_MUSTLOCK, does that cover the returned surfaces from SDL_ConvertSurface?

I will update the code according to this.

slouken:
> @slouken Just for clarity, SDL_ConvertSurface returns a new surface entirely right? Since a newly created surface will not be evaluated to true within SDL_MUSTLOCK, does that cover the returned surfaces from SDL_ConvertSurface?

Yes, it does. The only time MUSTLOCK returns true is if a surface has been RLE encoded for speed by the application.

LordOfXen:
Just a note, I think we should rename it back to SDL_IMAGE_SAVE_AVIF, because this was there before this addition, I did not add it _newly_.

Renaming SDL_IMAGE_SAVE_AVIF to SAVE_AVIF caused modification to existing code.

slouken:
> Just a note, I think we should rename it back to SDL_IMAGE_SAVE_AVIF, because this was there before this addition, I did not add it _newly_.
> 
> Renaming SDL_IMAGE_SAVE_AVIF to SAVE_AVIF caused modification to existing code.

Then go ahead and leave it as-is. We can make this more consistent in a separate PR.

sezero:
Fix type redefinition errors:

```diff
diff --git a/src/IMG_avif.c b/src/IMG_avif.c
index 973458e..ab9f28e 100644
--- a/src/IMG_avif.c
+++ b/src/IMG_avif.c
@@ -1073,9 +1073,9 @@ IMG_Animation* IMG_LoadAVIFAnimation_IO(SDL_IOStream* src)
 #if SAVE_AVIF
 
-typedef struct IMG_AnimationStreamContext
+struct IMG_AnimationStreamContext
 {
     avifEncoder *encoder;
     bool first_frame_added;
-} IMG_AnimationStreamContext;
+};
 
 static bool AnimationStream_AddFrame(struct IMG_AnimationStream *stream, SDL_Surface *surface, Uint64 pts)
```

LordOfXen:
Latest changes include the type redefinition fix sezero recommended, as well as eliminating the usage of RGB24 and instead convert/create all surfaces with RGBA32 where possible.

src/IMG_avif.c
```diff
@@ -756,3 +757,728 @@ bool IMG_SaveAVIF_IO(SDL_Surface *surface, SDL_IOStream *dst, bool closeio, int
...
+    if (!stream->ctx->encoder) {
+        SDL_free(stream->ctx);
+        stream->ctx = NULL;
+        return SDL_SetError("Couldn't create AVIF encoder");
+    }
+
+    int availableLCores = SDL_GetNumLogicalCPUCores();
+    int mThreads = (int)SDL_GetNumberProperty(props, "maxthreads", availableLCores / 2);
+    if (mThreads < 1 || mThreads > availableLCores)
+    {
```
slouken:
```suggestion
    if (mThreads < 1 || mThreads > availableLCores) {
```

slouken: This should clamp rather than return an error. The caller may not know how many logical cores there are (or have bothered to check)

LordOfXen: Yeah sure.

LordOfXen: I will change that in a second.

LordOfXen: Handled!

PR #595, opened 2025-08-06
Fix Octree Conditions in IMG_gif.c

LordOfXen:
A small bug fix to change `defined(SAVE_GIF)` to `SAVE_GIF` so the value of `SAVE_GIF` is considered while defining `SAVE_GIF_OCTREE`.

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -33,8 +33,8 @@
 #if SAVE_GIF
     #ifndef SAVE_GIF_OCTREE
         #define SAVE_GIF_OCTREE 1
-    #endif /* SAVE_GIF_OCTREE */
-#endif /* SAVE_GIF */
+    #endif
+#endif
 
 #ifdef LOAD_GIF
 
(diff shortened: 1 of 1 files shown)


src/IMG_gif.c
```diff
@@ -30,9 +30,11 @@
...
 
 // By default, non-indexed surfaces will be converted to indexed pixels using octree quantization.
-#if defined(SAVE_GIF)
-#define SAVE_GIF_OCTREE 1
-#endif
+#if SAVE_GIF
+    #ifndef SAVE_GIF_OCTREE
+        #define SAVE_GIF_OCTREE 1
+    #endif /* SAVE_GIF_OCTREE */
+#endif /* SAVE_GIF */
```
sezero: `#endif` comments serve nothing but hurting readers' eyes when they are this much close to their parent `#if`s

src/IMG_gif.c
```diff
@@ -1155,7 +1157,9 @@ static int lzwCompress(const uint8_t *indexedPixels, uint16_t width, uint16_t he
 }
 
 #if SAVE_GIF_OCTREE
-#define OCTREE_MAX_LEVELS 8
+#ifndef OCTREE_MAX_LEVELS
+    #define OCTREE_MAX_LEVELS 8
+#endif /* OCTREE_MAX_LEVELS */
```
sezero: same as above

LordOfXen: Removed from both.

Issue #599, opened 2025-08-07
IMG_AnimationStream Recommendation


LordOfXen:
I've run into a few performance issues with our current implementation of animation loading.
Since now we support loading WEBP, APNG, GIF, and AVIFS as animated, I made a test using WEBP, APNG, and AVIFS.
Currently, the animation is *entirely* loaded into IMG_Animation, resulting in heavy memory overhead, such as 600 megabytes for 320 framed WEBP, where each frame is RGBA8888. 

This is somewhat problematic; no one would want a single animated image to take up that much memory space.
The recommendation is below to solve this and #580.
Since the latest changes with the new IMG_AnimationStream haven't been released yet:

- Rename IMG_AnimationStream for encoding to something like IMG_AnimationEncodingStream.
- Have a secondary interface named IMG_AnimationDecodingStream.
- Instead of AddFrame, there would be GetFrames, where users can ask for frames sequentially.
- Update the body of animation loading functions in each format to use this internally, where the GetFrames will be given the entire number of frames available for full loading.

This results in a smoother experience with loading animations, loading frames on demand without hassling the memory on the user's computer.
This also results in new opportunities to be considered while implementing, E.G. #580.

Please let me know your thoughts.

slouken:
This sounds like a good idea. The existing anim loading API could be implemented on top of this new interface as well.

LordOfXen:
> This sounds like a good idea. The existing anim loading API could be implemented on top of this new interface as well.

Yes, that's what I meant.
We can also expose a PropertiesID as well for format specific settings, like to get how many times to repeat etc. as mentioned in #580.

So, this can solve many things.

LordOfXen:
`GetFrames` would return an integer similar to IO functions, stating how many frames have been loaded.
Like at the end of animation where 3 frames are left but the user asked to load 5 frames, the return value would state 3, or 0 (when no frames are left), indicating the end of the animation stream.

To get the loaded SDL_Surface array, that'd be passed by reference from parameters or vice versa, an integer to be passed through parameters, and an array to be returned instead.

LordOfXen:
But we'd also need a `Reset` function to rewind the stream (Animation Stream) to the beginning for looping, so a continuous on-demand animation can be achieved.

LordOfXen:
I am coding the changes right now, but I will need support when I merge here.

LordOfXen:
[...]

PR #600, opened 2025-08-09
Heavy Changes to Animation Stream & Animation Loading

LordOfXen:
Introducing the changes mentioned in #599 in most possible efficient ways.

### Changes
- Verify all calls to SDL functions returns successful, the previous implementation in GIF and PNG wasn't checking whether `SDL_BlitSurface` and such's return value or not.
- Rename `IMG_AnimationStream` to `IMG_AnimationEncoderStream`.
- Rename all pre-defined settings for `IMG_AnimationEncoderStream` to include `encoder` prefix.
- Introduce `IMG_AnimationDecoderStream`.
- Introduce similar pre-defined settings for `IMG_AnimationDecoderStream`.
- Change existing animation loading API to use IMG_AnimationDecoderStream in the background. This has been done smoothly with a single call to `IMG_DecodeAsAnimation` for all formats.

### Planned
- Review the code for bugs.
- Review the changes to ensure it will behave the same as the previous implementation.
- Review encoder-specific settings for exposed API to include defines for most standardized ones (e.g. loop count).

### Actual Design
`IMG_CreateAnimationDecoderStreamWithProperties` to be called with `SDL_PropertiesID` for both inputs and outputs. This means the decoder stream can be used to extract data, and user should not be forced to load any frames at all. The existing implementation supports this. You only create the decoder stream, and then check your `SDL_PropertiesID` with filled data, then free it before getting any frames.

`GetFrames` works with the new struct `IMG_AnimationDecoderFrames` for holding the frame data, this has been created to isolate from the existing `IMG_Animation` interface completely and to hold `PTS` instead of integer milliseconds for flexibility.
The current `IMG_AnimationDecoderFrames` holds `PTS` in the desired `timebase_numerator` and `timebase_denominator`, no precision will be lost for sensitive animation decoders.

`Reset` works by rewinding the underlying iterator/SDL_IOStream/DecoderContext depending on the format and resets all frame indexing to 0 or -1 for re-reading from the beginning. The naming kept simple but can be changed if not met with the requirementes of the existing API. It's purpose is to be able to re-use the existing decoder stream for looping without having to destroy and re-create the entire context. **This (Reset function/feature) remains untested as of (8/8/2025 06:33 PM).**

Other functions `Close` and `IMG_CreateAnimationDecoderStreamWithProperties` are identical to our encoder counterparts.
Please feel free to recommend changes as needed, and make sure to review the changes carefully to make sure no mistakes are made.

### DIFF
--- a/VisualC/SDL_image.vcxproj
+++ b/VisualC/SDL_image.vcxproj
@@ -26,6 +26,7 @@
     <ProjectGuid>{2BD5534E-00E2-4BEA-AC96-D9A92EA24696}</ProjectGuid>
     <RootNamespace>SDL3_image</RootNamespace>
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">10.0</VisualStudioVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
@@ -38,7 +39,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset Condition="'$(VisualStudioVersion)' != '10.0'">$(DefaultPlatformToolset)</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' != '10.0'">ClangCL</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
@@ -180,7 +181,7 @@
       </HeaderFileName>
     </Midl>
     <ClCompile>
-      <AdditionalIncludeDirectories>external\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>C:\Users\emirm\Downloads\SDL3-3.2.16\include;external\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>DLL_EXPORT;NDEBUG;WIN32;_WINDOWS;USE_STBIMAGE;LOAD_AVIF;LOAD_AVIF_DYNAMIC="libavif-16.dll";LOAD_BMP;LOAD_GIF;LOAD_JPG;LOAD_LBM;LOAD_PCX;LOAD_PNG;LOAD_PNM;LOAD_QOI;LOAD_SVG;LOAD_TGA;LOAD_TIF;LOAD_TIF_DYNAMIC="libtiff-6.dll";LOAD_WEBP;LOAD_WEBP_DYNAMIC="libwebp-7.dll";LOAD_WEBPDEMUX_DYNAMIC="libwebpdemux-2.dll";LOAD_WEBPMUX_DYNAMIC="libwebpmux-3.dll";LOAD_LIBPNG_DYNAMIC="libpng16-16.dll";SDL_IMAGE_LIBPNG;LOAD_XCF;LOAD_XPM;LOAD_XV;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
       <WarningLevel>Level3</WarningLevel>
@@ -193,12 +194,17 @@
       <AdditionalDependencies>SDL3.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <OptimizeReferences>true</OptimizeReferences>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
-      <AdditionalLibraryDirectories>C:\Users\emirm\Downloads\SDL_image-09e29993126b13c3dd589c5f7a5806c09b43279f\libs;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>C:\Users\emirm\Downloads\SDL3-3.2.16\VisualC\x64\Release;C:\Users\emirm\Downloads\SDL_image-09e29993126b13c3dd589c5f7a5806c09b43279f\libs;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
+    <PostBuildEvent>
+      <Command>xcopy "C:\Users\emirm\OneDrive\Belgeler\sdl_image_workon\SDL_image\VisualC\x64\Release\SDL3_image.dll" "C:\Users\emirm\source\repos\SDL3_image_testbed\x64\Debug\" /I /Y
+xcopy "C:\Users\emirm\OneDrive\Belgeler\sdl_image_workon\SDL_image\VisualC\x64\Release\SDL3_image.pdb" "C:\Users\emirm\source\repos\SDL3_image_testbed\x64\Debug\" /I /Y</Command>
+    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="..\src\IMG.c" />
-    <ClCompile Include="..\src\IMG_anim.c" />
+    <ClCompile Include="..\src\IMG_anim_decoder.c" />
+    <ClCompile Include="..\src\IMG_anim_encoder.c" />
     <ClCompile Include="..\src\IMG_avif.c" />
     <ClCompile Include="..\src\IMG_bmp.c" />
     <ClCompile Include="..\src\IMG_gif.c" />
@@ -225,6 +231,8 @@
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\include\SDL3_image\SDL_image.h" />
+    <ClInclude Include="..\src\IMG_anim_decoder.h" />
+    <ClInclude Include="..\src\IMG_anim_encoder.h" />
     <ClInclude Include="..\src\IMG_libpng.h" />
     <ClInclude Include="..\src\IMG_gif.h" />
     <ClInclude Include="..\src\IMG_avif.h" />
@@ -435,5 +443,3 @@
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
 </Project>
-
-
--- a/VisualC/SDL_image.vcxproj
+++ b/VisualC/SDL_image.vcxproj
@@ -181,7 +181,7 @@
       </HeaderFileName>
     </Midl>
     <ClCompile>
-      <AdditionalIncludeDirectories>C:\Users\emirm\Downloads\SDL3-3.2.16\include;external\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>external\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>DLL_EXPORT;NDEBUG;WIN32;_WINDOWS;USE_STBIMAGE;LOAD_AVIF;LOAD_AVIF_DYNAMIC="libavif-16.dll";LOAD_BMP;LOAD_GIF;LOAD_JPG;LOAD_LBM;LOAD_PCX;LOAD_PNG;LOAD_PNM;LOAD_QOI;LOAD_SVG;LOAD_TGA;LOAD_TIF;LOAD_TIF_DYNAMIC="libtiff-6.dll";LOAD_WEBP;LOAD_WEBP_DYNAMIC="libwebp-7.dll";LOAD_WEBPDEMUX_DYNAMIC="libwebpdemux-2.dll";LOAD_WEBPMUX_DYNAMIC="libwebpmux-3.dll";LOAD_LIBPNG_DYNAMIC="libpng16-16.dll";SDL_IMAGE_LIBPNG;LOAD_XCF;LOAD_XPM;LOAD_XV;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
       <WarningLevel>Level3</WarningLevel>
@@ -194,12 +194,7 @@
       <AdditionalDependencies>SDL3.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <OptimizeReferences>true</OptimizeReferences>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
-      <AdditionalLibraryDirectories>C:\Users\emirm\Downloads\SDL3-3.2.16\VisualC\x64\Release;C:\Users\emirm\Downloads\SDL_image-09e29993126b13c3dd589c5f7a5806c09b43279f\libs;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
-    <PostBuildEvent>
-      <Command>xcopy "C:\Users\emirm\OneDrive\Belgeler\sdl_image_workon\SDL_image\VisualC\x64\Release\SDL3_image.dll" "C:\Users\emirm\source\repos\SDL3_image_testbed\x64\Debug\" /I /Y
-xcopy "C:\Users\emirm\OneDrive\Belgeler\sdl_image_workon\SDL_image\VisualC\x64\Release\SDL3_image.pdb" "C:\Users\emirm\source\repos\SDL3_image_testbed\x64\Debug\" /I /Y</Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="..\src\IMG.c" />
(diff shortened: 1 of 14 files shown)


src/IMG_webp.c
```diff
@@ -543,7 +543,7 @@ bool IMG_CreateWEBPAnimationDecoderStream(IMG_AnimationDecoderStream *stream, SD
 
     Sint64 stream_size = SDL_GetIOSize(stream->src);
     if (stream_size <= 0) {
-        SDL_SetError("Stream has no data (size: %lld)", stream_size);
+        SDL_SetError("Stream has no data (size: %lld)", (long long)stream_size);
```
sezero: `SDL_SetError("Stream has no data (size: %" SDL_PRIs64 ")", stream_size);`

LordOfXen: Replaced, thanks.

LordOfXen:
~Please do not merge this yet even if all tests passes, I need to make sure it behaves as before and works as expected for all of the changes first.~
Edit: Tests completed!

CMakeLists.txt
```diff
@@ -260,7 +260,8 @@ set(BUILD_SHARED_LIBS ${SDLIMAGE_BUILD_SHARED_LIBS})
 add_library(${sdl3_image_target_name}
     src/IMG.c
     src/IMG_WIC.c
-    src/IMG_anim.c
+    src/IMG_anim_encoder.c
+    src/IMG_anim_decoder.c
```
sezero, line 264: do the same in Android.mk

Green-Sky (commented):
Has there gone any thought in merging the two stream types?
Like it is with `SDL_IOStream` ? Or does that have too many downsides?

LordOfXen:
> Has there gone any thought in merging the two stream types? Like it is with `SDL_IOStream` ? Or does that have too many downsides?

It does not have downsides as far as I can think right now but that requires us to have all input and output functions to be in the same struct as well, like add frame and get frames in the same place logically doesn't fit together in my opinion.

However, it can still be made with modes or checks for nulls, the only downgrade I can see is flexibility in the future in case we want a decoder or encoder only feature (for example).

LordOfXen:
Note that merging those two would also result in complexity and less readability, maintaining a single interface containing both jobs is harder than maintaining two different interfaces indicating their jobs explicitly.

slouken:
> Has there gone any thought in merging the two stream types? Like it is with `SDL_IOStream` ? Or does that have too many downsides?

Yeah, there's no benefit to having a single stream type that does both encoding and decoding. It's error-prone for the user and adds confusion for the implementation.

LordOfXen:
All of the tests mentioned in the header have been successfully completed, and all features seem to work as expected as of now.

slouken:
I suggest removing "stream" from the naming, so we have AnimationEncoder and AnimationDecoder interfaces.

LordOfXen:
> I suggest removing "stream" from the naming, so we have AnimationEncoder and AnimationDecoder interfaces.

That'd actually be better for clarity.

LordOfXen:
I'll dig in to remove all "stream"'s from names if OK for everyone.

slouken:
> I'll dig in to remove all "stream"'s from names if OK for everyone.

Yes please, this is a great time to do it.

LordOfXen:
> > I'll dig in to remove all "stream"'s from names if OK for everyone.
> 
> Yes please, this is a great time to do it.

On it!

VisualC/SDL_image.vcxproj
```diff
@@ -26,6 +26,7 @@
     <ProjectGuid>{2BD5534E-00E2-4BEA-AC96-D9A92EA24696}</ProjectGuid>
     <RootNamespace>SDL3_image</RootNamespace>
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">10.0</VisualStudioVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
slouken:
```suggestion
```

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+ * \sa IMG_CreateAnimationEncoderWithProperties
+ * \sa IMG_AddAnimationEncoderFrame
+ * \sa IMG_CloseAnimationEncoder
  */
-extern SDL_DECLSPEC IMG_AnimationStream * SDLCALL IMG_CreateAnimationStream(const char *file);
+extern SDL_DECLSPEC IMG_AnimationEncoder * SDLCALL IMG_CreateAnimationEncoder(const char *file);
 
 /**
- * Create an animation stream and save it to an IOStream.
+ * Create an animation encoder and save it to an IOStream.
```
```suggestion
 * Create an encoder to save a series of images to an IOStream.
```
This doesn't make sense... maybe this instead?

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
 /**
- * An object representing a stream of images being saved.
+ * An object representing the encoder context.
  */
-typedef struct IMG_AnimationStream IMG_AnimationStream;
+typedef struct IMG_AnimationEncoder IMG_AnimationEncoder;
 
 /**
- * Create an animation stream and save it to a file.
+ * Create an animation encoder and save it to a file.
```
```suggestion
 * Create an encoder to save a series of images to a file.
```
This doesn't make sense... maybe this instead?

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+{
+    int w;                  /**< The width of the frames */
+    int h;                  /**< The height of the frames */
+    int count;              /**< The number of frames */
+    SDL_Surface **frames;   /**< An array of frames */
+    Sint64 *delays;            /**< An array of frame delays */
+} IMG_AnimationDecoderFrames;
+
+/**
+ * Create an animation decoder and save it to a file.
```
```suggestion
 * Create a decoder to read a series of images from a file.
```

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+ *
+ * The file type is determined from the file extension, e.g. "file.webp" will
+ * be encoded using WEBP.
```
```suggestion
 * be decoded using WEBP.
```

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+ * \since This function is available since SDL_image 3.4.0.
+ *
+ * \sa IMG_CreateAnimationDecoder_IO
+ * \sa IMG_CreateAnimationDecoderWithProperties
+ * \sa IMG_CloseAnimationDecoder
+ */
+extern SDL_DECLSPEC IMG_AnimationDecoder * SDLCALL IMG_CreateAnimationDecoder(const char *file);
+
+/**
+ * Create an animation decoder and save it to an IOStream.
```
```suggestion
 * Create a decoder to read a series of images from an IOStream.
```

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+ * \param decoderFrames an instance of IMG_AnimationDecoderFrames to be filled with
+ *        loaded frames and their delays.
+ * \returns true on success or false on failure; call SDL_GetError() for more
+ *          information.
+ *
+ * \since This function is available since SDL_image 3.4.0.
+ *
+ * \sa IMG_ResetAnimationDecoder
+ */
+extern SDL_DECLSPEC bool SDLCALL IMG_GetAnimationDecoderFrames(IMG_AnimationDecoder *decoder, int framesToLoad, IMG_AnimationDecoderFrames* decoderFrames);
```
slouken: This function should probably return the next frame and PTS rather than a structure of them. If someone wants more than one frame they can call it multiple times.

We probably also need a function to return the number of frames available.

include/SDL3_image/SDL_image.h
```diff
@@ -2404,140 +2404,332 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+ *
+ * \param decoder the animation decoder.
+ * \param pts the presentation timestamp to convert.
+ * \returns the presentation timestamp in milliseconds.
+ *
+ * \since This function is available since SDL_image 3.4.0.
+ *
+ * \sa IMG_GetAnimationDecoderFrames
+ */
+extern SDL_DECLSPEC int SDLCALL IMG_GetAnimationDecoderPresentationTimestampMS(IMG_AnimationDecoder *decoder, Sint64 pts);
```
slouken, line 2723: We probably don't need this function, see above.

include/SDL3_image/SDL_image.h
LordOfXen: Yeah sure that can be implemented too, but both have disadvantages and advantages.

We do not need a function for returning the number of frames available, the decoder can write inside the given SDL_PropertiesID, maybe we can just hold those there?

User can call create with properties then after the call to create, they can check the properties filled with whatever they need from the decoder, including decoder specific settings like loop count.

LordOfXen: We can *additionally* add IMG_GetNextAnimationDecoderFrame over the existing IMG_GetAnimationDecoderFrames, in which case IMG_GetAnimationDecoderFrames would use IMG_GetNextAnimationDecoderFrame to try getting as many frames as requested or as many as available, whichever is less.

include/SDL3_image/SDL_image.h
LordOfXen: This just uses timebase_numerator and timebase_denominator to convert PTS to an actual millisecond representation, as done in IMG_DecodeAsAnimation.

The user can also do that calculation themselves as well, unless you mean something else.

include/SDL3_image/SDL_image.h
slouken: The create properties should be considered read-only. Instead we should have an explicit function to get the number of frames. We’ll probably _also_ add a function to get codec specific properties, but that can be added later.

slouken: We should have just one function to get a frame, and it can match the encoder function for symmetry, using out pointers. I don’t think we want to complicate the interface with additional functions. We can add those later if we need to.

slouken: If you want we can merge what you have and I can clean up and simplify the interface.

LordOfXen: Okay. I can lastly change the IMG_GetAnimationDecoderFrames with IMG_GetNextAnimationDecoderFrame, then after that we can merge this.

But in that case the IMG_AnimationDecoderFrames becomes useless so should be removed as well, we'd return SDL_Surface and ask for a Sint64 in the parameters to assign the pts to in IMG_GetNextAnimationDecoderFrame. Or if you have something else in mind, please feel free to explain it.

Also is naming for IMG_GetNextAnimationDecoderFrame good?
Please tell me those so no misunderstandings happens while making the changes.

LordOfXen: Name can also be IMG_DecodeNextFrame.

LordOfXen: Just noting that my next commit would be probably after 8-10 hours, if you have free time to work on this in earlier time than that, feel free to merge this and make those changes 🙂

I don't want to push a lot on you so if you have much to care of, leave it to me I will work on it as soon as possible.

LordOfXen:
I've made the necessary changes as mentioned earlier by @slouken, but I've had some issues that I needed to decide about to fix.
First, I'll list the changes I have made:
- Removal of unnecessary code left from initial tests/implementation.
- Bug fixes.
- Change `IMG_GetAnimationDecoderFrames` to `IMG_GetNextAnimationDecoderFrame` for decoding a single frame per call.
- Removal of `IMG_AnimationDecoderFrames` as it is no longer needed.
- Introduce `IMG_GetAnimationDecoderMetadata` that returns a `SDL_PropertiesID` from a specific `IMG_AnimationDecoder`. This is set when creating the decoder and is destroyed when the decoder is closed. Users should **NEVER** rely on the data returned by this metadata, as it may not contain the frame count for all decoders. For example, GIF decoders do not contain a frame count in their header or binary structure; they simply want you to continuously decode until you hit the GIF trailer and increase the count each time you successfully decode a frame before hitting the trailer. That'd require us to decode the entire GIF first to get a dummy count, then assign it and reset the decoder, which is expensive in terms of this implementation for big animations (we can also loop through the binary without decompressing frames, but that's a whole different job that should be considered after this PR I suggest). We instead rely on the behaviour of `IMG_GetNextAnimationDecoderFrame` to give us a `NULL` pointer for the surface, indicating the end of data for that decoding process.
- Introduce `IMG_PROP_ANIMATION_DECODER_METADATA_FRAME_COUNT_NUMBER` define.
- Introduce `IMG_PROP_ANIMATION_DECODER_METADATA_LOOP_COUNT_NUMBER` define.

All of the underlying decoder codes have been revamped according to the changes listed above.
The first decision I had to made was the return value of `IMG_GetNextAnimationDecoderFrame`, either to be bool or SDL_Surface*, first I made it `SDL_Surface*` to return `NULL` when we no longer have more frames to decode, but this was complicating the situation in case an error occurs, the user would have to first clear previous error before calling `IMG_GetNextAnimationDecoderFrame`, then call it and then check the returned `SDL_Surface*` if its `NULL` or not, if it is `NULL`, check `SDL_GetError` if it contains any error, if it does, it failed, if it doesn't, we just have no more frames.

That step is quite complicated, and can be misinterpreted sometimes, or cause forgotten steps, like forgetting to call `SDL_ClearError` will cause users to believe the decoding failed even if it didn't.

For this reason, I changed the return value of `IMG_GetNextAnimationDecoderFrame` from `SDL_Surface*` to `bool`, and included `SDL_Surface**` frame in the parameters of `IMG_GetNextAnimationDecoderFrame`.

This way it returns true if `IMG_GetNextAnimationDecoderFrame` succeeds, but if there are no more frames, you get a `NULL` from the frame parameter.
This behaviour is much better than the previous one returning `SDL_Surface*` in my opinion, and I decided to keep it this way.

LordOfXen:
Note that instead of taking `SDL_Surface**`, we can also make a struct like `IMG_DecodedFrame` and take the address of that to fill with `PTS` and `SDL_Surface*`.

Additionally, we can rename `IMG_GetNextAnimationDecoderFrame` as well, to something like `IMG_DecodeNextFrame`, but I am not sure.

Or we can also leave both as is, just let me know what you think.

LordOfXen:
### More suggestions
`IMG_CloseAnimationDecoder`/`IMG_CloseAnimationEncoder`: Maybe we should consider the naming for those two, because they don't only "Close", they do last steps for encoding/decoding as well, such as assembling part for WEBP.

Does `FinalizeAnimationEncoder` fit more for the Encoder part?

After a close review of each decoder, the decoding part for close only destroys the initialized data and interfaces for the underlying decoder API; therefore, it barely does anything regarding *finalizing* or *closing* except when `closeio` is true. Maybe it should be named `DestroyAnimationDecoder`?

Or we can rename both to `DestroyAnimationDecoder`/`DestroyAnimationEncoder` and add one more function to the encoder called ``FinalizeEncoder``.

Again, this is just for clarity, but worth considering before we merge this.

LordOfXen:
FYI, I've redone the tests mentioned in the header and they continue to work as before after the change from `GetFrames` to `GetNextFrame`.

slouken:
> For this reason, I changed the return value of `IMG_GetNextAnimationDecoderFrame` from `SDL_Surface*` to `bool`, and included `SDL_Surface**` frame in the parameters of `IMG_GetNextAnimationDecoderFrame`.

This is exactly what I was thinking as well.

slouken:
> ### More suggestions
> `IMG_CloseAnimationDecoder`/`IMG_CloseAnimationEncoder`: Maybe we should consider the naming for those two, because they don't only "Close", they do last steps for encoding/decoding as well, such as assembling part for WEBP.
> 
> Does `FinalizeAnimationEncoder` fit more for the Encoder part?
> 
> After a close review of each decoder, the decoding part for close only destroys the initialized data and interfaces for the underlying decoder API; therefore, it barely does anything regarding _finalizing_ or _closing_ except when `closeio` is true. Maybe it should be named `DestroyAnimationDecoder`?
> 
> Or we can rename both to `DestroyAnimationDecoder`/`DestroyAnimationEncoder` and add one more function to the encoder called `FinalizeEncoder`.
> 
> Again, this is just for clarity, but worth considering before we merge this.

Having a separate finalize step is easy to forget, so I prefer to have a bool Close function at least for encoding. The SDL Destroy functions return void and just do teardown, so I don't think those are appropriate here. I like the encoding and decoding API having symmetry, so I think what you have here is fine.

slouken:
I'll go ahead and merge what you have and iterate on it, thanks!

LordOfXen:
> Having a separate finalize step is easy to forget, so I prefer to have a bool Close function at least for encoding. The SDL Destroy functions return void and just do teardown, so I don't think those are appropriate here. I like the encoding and decoding API having symmetry, so I think what you have here is fine.

Yeah, well, forgetting is an acceptable reason for this scenario, as well as inconsistency between encoding and decoding.
I've stated all the ones I had in mind for review anyway.

LordOfXen:
> I'll go ahead and merge what you have and iterate on it, thanks!

Yeah sure, my pleasure!

sezero:
Guys, be more careful please? https://github.com/libsdl-org/SDL_image/commit/6d8fe18ff227cf071a44622e022499cdae9f2500

LordOfXen:
> Guys, be more careful please? [6d8fe18](https://github.com/libsdl-org/SDL_image/commit/6d8fe18ff227cf071a44622e022499cdae9f2500)

Sorry it probably slipped.

PR #601, opened 2025-08-10
Cleanup and temporarily removed decoder metadata

slouken:


slouken:
@LordOfXen, I temporarily removed the decoder metadata so we can do a more complete pass on metadata across all interfaces. Can you review the rest of this as well?

src/SDL_image.sym
```diff
@@ -63,26 +63,22 @@ SDL3_image_0.0.0 {
     IMG_SaveTGA_IO;
     IMG_SaveGIF;
     IMG_SaveGIF_IO;
-    IMG_SaveWEBP;
-    IMG_SaveWEBP_IO;
```
sezero, line 67: Why did you remove these two? Bad rebase after https://github.com/libsdl-org/SDL_image/commit/6d8fe18ff227cf071a44622e022499cdae9f2500 ??

sezero: .. and I hope there aren't any other badnesses after rebasing after https://github.com/libsdl-org/SDL_image/commit/6d8fe18ff227cf071a44622e022499cdae9f2500

slouken: Yep.

src/IMG_anim_decoder.c
```diff
@@ -238,164 +211,104 @@ IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int
...
-            for (int i = 0; i < actualCount; ++i) {
-                SDL_DestroySurface(frames[i]);
-            }
-
-            SDL_free(frames);
-            SDL_free(delays);
-            IMG_CloseAnimationDecoder(decoder);
-            return NULL;
-        }
-        delays = tempDelays;
```
LordOfXen, line 332: Note that removal of resizing/fitting the arrays after the loop may cause unnecessary allocation for big animations.
Assume 2048 as the actual, one more addition will trigger resize, it will grow up to 4096, just for 2049 items.

May not be a big concern, but making sure it's on the mind.

slouken: We size anim->frames and anim->delays to the exact size we want, and then free frames and delays before returning, so there's no need to do an additional resizing down.

LordOfXen:
The rest mostly looks okay, but I need to test it to make sure. Give me a minute to run this through the tests I've made in the actual pull request.

src/IMG_anim_decoder.c
LordOfXen: Ah right! Since we free the temporary arrays either way before exiting the function, good catch!

LordOfXen:
Tested AVIF, GIF, WEBP, APNG, all work as before.

PR #603, opened 2025-08-11
Animation Encoding/Cecoding Metadata (aka PROPS) Feature

LordOfXen:
Introduces the new support to get (decoder) or set (encoder) metadata.

### Added Features
- Added `xmlman.c` and `xmlman.h`, which is simple without any dependency, written by me (no library or license), it simply parses data from XMP and writes data back to XMP from hard-coded XML structures. 
- Added `IMG_PROP_ANIMATION_IGNORE_PROPS_BOOLEAN` for ignoring props in encoder and decoder; some users may want to skip this process entirely. This is what they'd look for.
- Added `IMG_PROP_ANIMATION_DESCRIPTION_STRING` for reading or writing description metadata from decoders and to encoders.
- Added `IMG_PROP_ANIMATION_COPYRIGHT_STRING` for reading or writing copyright metadata from decoders and to encoders.
- Added `IMG_PROP_ANIMATION_TITLE_STRING` for reading or writing copyright metadata from decoders and to encoders.
- Added `IMG_PROP_ANIMATION_AUTHOR_STRING` for reading or writing author/creator metadata from decoders and to encoders
- Added `IMG_PROP_ANIMATION_CREATION_TIME_STRING` for reading or writing image creation time metadata from decoders and to encoders. Some images do not have to be created at the time SDL encodes this; therefore, I exposed it to be set by users if desired as well.
- Added `IMG_PROP_ANIMATION_LOOP_COUNT_NUMBER` for reading or writing the desired loop count for the animation.

### Currently Supported Formats
- WEBP supports all of the well-defined metadata for both decoding and encoding.
- AVIF supports all of the well-defined metadata for both decoding and encoding.
- PNG supports all of the well-defined metadata for both decoding and encoding.
- GIF only supports comment/description and loop count (from the NETSCAPE 2.0 extension) metadata since it's an old format with limited features.

### Tests
Tests still continue, WEBP and AVIF work as expected; however, GIF and PNG remain untested as of 8/11/2025 - 10:04 AM.

Feel free to make recommendations as well as bug fixes.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -75,6 +75,7 @@ LOCAL_SRC_FILES :=  		\
     src/IMG.c           	\
     src/IMG_anim_encoder.c      \
     src/IMG_anim_decoder.c      \
+    src/xmlman.c      		\
     src/IMG_avif.c      	\
     src/IMG_bmp.c       	\
     src/IMG_gif.c       	\
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,7 @@ add_library(${sdl3_image_target_name}
     src/IMG_WIC.c
     src/IMG_anim_encoder.c
     src/IMG_anim_decoder.c
+    src/xmlman.c
     src/IMG_avif.c
     src/IMG_bmp.c
     src/IMG_gif.c
--- a/include/SDL3_image/SDL_image.h
+++ b/include/SDL3_image/SDL_image.h
@@ -2403,6 +2403,14 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
  */
 extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadWEBPAnimation_IO(SDL_IOStream *src);
 
+#define IMG_PROP_ANIMATION_IGNORE_PROPS_BOOLEAN                 "SDL_image.animation.ignore_props"
+#define IMG_PROP_ANIMATION_DESCRIPTION_STRING                   "SDL_image.animation.description"
+#define IMG_PROP_ANIMATION_COPYRIGHT_STRING                     "SDL_image.animation.copyright"
+#define IMG_PROP_ANIMATION_TITLE_STRING                         "SDL_image.animation.title"
+#define IMG_PROP_ANIMATION_AUTHOR_STRING                        "SDL_image.animation.author"
+#define IMG_PROP_ANIMATION_CREATION_TIME_STRING                 "SDL_image.animation.creation_time"
+#define IMG_PROP_ANIMATION_LOOP_COUNT_NUMBER                    "SDL_image.animation.loop_count"
+
 /**
  * An object representing the encoder context.
  */
@@ -2627,6 +2635,26 @@ extern SDL_DECLSPEC IMG_AnimationDecoder * SDLCALL IMG_CreateAnimationDecoderWit
 #define IMG_PROP_ANIMATION_DECODER_CREATE_TYPE_STRING                    "SDL_image.animation_decoder.create.type"
 #define IMG_PROP_ANIMATION_DECODER_CREATE_TIMEBASE_NUMERATOR_NUMBER      "SDL_image.animation_decoder.create.timebase.numerator"
 #define IMG_PROP_ANIMATION_DECODER_CREATE_TIMEBASE_DENOMINATOR_NUMBER    "SDL_image.animation_decoder.create.timebase.denominator"
+
+
+/**
+ * Get the properties of an animation decoder.
+ *
+ * This function returns the properties of the animation decoder, which
+ * holds information about the underlying image such as description,
+ * copyright text and loop count.
+ *
+ * \param decoder the animation decoder.
+ * \returns the properties ID of the animation decoder, or 0 if there are no
+ *          properties; call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL_image 3.4.0.
+ *
+ * \sa IMG_CreateAnimationDecoder
+ * \sa IMG_CreateAnimationDecoder_IO
+ * \sa IMG_CreateAnimationDecoderWithProperties
+ */
+extern SDL_DECLSPEC SDL_PropertiesID SDLCALL IMG_GetAnimationDecoderProperties(IMG_AnimationDecoder* decoder);
 
 /**
  * Get the next frame in an animation decoder.
--- a/src/IMG_anim_decoder.h
+++ b/src/IMG_anim_decoder.h
@@ -34,6 +34,7 @@ struct IMG_AnimationDecoder
     bool (*Close)(IMG_AnimationDecoder *decoder);
 
     IMG_AnimationDecoderContext *ctx;
+    SDL_PropertiesID props;
 };
 
 extern IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int maxFrames);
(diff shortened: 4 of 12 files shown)


include/SDL3_image/SDL_image.h
```diff
@@ -2403,6 +2403,14 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
  */
 extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadWEBPAnimation_IO(SDL_IOStream *src);
 
+#define IMG_PROP_ANIMATION_IGNORE_PROPS_BOOLEAN                 "SDL_image.animation.ignore_props"
+#define IMG_PROP_ANIMATION_DESCRIPTION_STRING                   "SDL_image.animation.description"
```
```suggestion
#define IMG_PROP_METADATA_DESCRIPTION_STRING                   "SDL_image.metadata.description"
```
This will eventually be supported on single-image load/save, so the naming should reflect that.

src/IMG_anim_decoder.h
```diff
@@ -34,6 +34,7 @@ struct IMG_AnimationDecoder
     bool (*Close)(IMG_AnimationDecoder *decoder);
 
     IMG_AnimationDecoderContext *ctx;
+    SDL_PropertiesID props;
```
slouken: Not super important, but I would move this up with the other data members above.

src/IMG_avif.c
```diff
@@ -1043,6 +1050,27 @@ static bool IMG_AnimationDecoderClose_Internal(IMG_AnimationDecoder* decoder)
...
+    if (title) {
+        SDL_free((void *)title);
+    }
+    if (creator) {
+        SDL_free((void *)creator);
+    }
+    if (createdate) {
+        SDL_free((void *)createdate);
+    }
+
```
slouken:
```suggestion
```
String properties are automatically cleaned up when the properties container is destroyed.

src/IMG_anim_decoder.h
LordOfXen: First I will check why gif fails, that's also the reason of build failures, then I will adjust the definitions with your recommendations.

src/IMG_gif.c
```diff
@@ -849,6 +939,20 @@ bool IMG_CreateGIFAnimationDecoder(IMG_AnimationDecoder* decoder, SDL_Properties
...
     decoder->Close = IMG_AnimationDecoderClose_Internal;
 
+    if (!IMG_AnimationDecoderGetGIFHeader(decoder)) {
+        return false;
+    }
+
+    bool ignoreProps = SDL_GetBooleanProperty(props, IMG_PROP_ANIMATION_IGNORE_PROPS_BOOLEAN, false);
+    if (!ignoreProps) {
+        // Set well-defined properties.
+        SDL_SetNumberProperty(props, IMG_PROP_ANIMATION_LOOP_COUNT_NUMBER, (Sint64)ctx->loop_count);
```
slouken: These should be set in decoder->props

LordOfXen:
@slouken Latest changes addresses all of the ones you mentioned plus the fix for gif extension parser and also forgotten parts to be changed from Sint64 for PTS to Uint64.

LordOfXen:
However, please do not merge this yet, as I haven't completely tested GIF and PNG yet.

Android.mk
```diff
@@ -75,6 +75,7 @@ LOCAL_SRC_FILES :=  		\
     src/IMG.c           	\
     src/IMG_anim_encoder.c      \
     src/IMG_anim_decoder.c      \
+    src/xmlman.c      		\
```
slouken: Minor nit, can you move this to the bottom?

CMakeLists.txt
```diff
@@ -262,6 +262,7 @@ add_library(${sdl3_image_target_name}
     src/IMG_WIC.c
     src/IMG_anim_encoder.c
     src/IMG_anim_decoder.c
+    src/xmlman.c
```
slouken: here as well?

include/SDL3_image/SDL_image.h
```diff
@@ -2403,6 +2403,14 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
  */
 extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadWEBPAnimation_IO(SDL_IOStream *src);
 
+#define IMG_PROP_METADATA_IGNORE_PROPS_BOOLEAN                 "SDL_image.metadata.ignore_props"
```
slouken: These should probably move up, maybe after IMG_Version?

slouken: Actually, corrected, probably down right after IMG_GetAnimationDecoderProperties(), that way they'll be grouped on the wiki.

LordOfXen: Yeah sure, since they are no longer related to animation anymore.

src/IMG_webp.c
```diff
@@ -867,9 +932,68 @@ static bool IMG_CloseWEBPAnimation(IMG_AnimationEncoder *encoder)
         goto done;
     }
 
-    if (!lib.WebPAnimEncoderAssemble(ctx->encoder, &data)) {
-        error = "WebPAnimEncoderAssemble() failed";
-        goto done;
+    if (!ctx->rights && !ctx->desc && ctx->loop_count == 0)
+    {
```
slouken:
```suggestion
    if (!ctx->rights && !ctx->desc && ctx->loop_count == 0) {
```

include/SDL3_image/SDL_image.h
LordOfXen: I just moved them up and saw your second comment. What's the latest thought on this?

LordOfXen: If we prefer sorting based on usage, the top location is better, if we gonna prefer sorting based on date added, then the previous was better, so it is up to you.

src/IMG_libpng.c
```diff
@@ -33,7 +33,6 @@
 
 #ifdef SDL_IMAGE_LIBPNG
 #include <png.h>
-#include <limits.h>
```
sezero, line 36: Do not remove this: it is needed for INT_MAX. Build fails against libpng12 without it.

LordOfXen: May I ask where INT_MAX usage is?

sezero: Here: https://github.com/libsdl-org/SDL_image/blob/73417c5567d7b6fadfdcc460108914b464c5bfa5/src/IMG_libpng.c#L1376

LordOfXen: We can just adjust that to SDL_MAX_SINT32 (I don't remember exact naming but you know what I mean).

I don't think we should include limits.h for this.

sezero: > I don't think we should include limits.h for this.

I don't see the harm, but ...

> We can just adjust that to SDL_MAX_SINT32

yes, `SDL_MAX_SINT32` is perfectly fine for it.

include/SDL3_image/SDL_image.h
slouken: Go ahead and move them after IMG_GetAnimationDecoderProperties(), that way they'll be associated with that function on the wiki. We don't have specific functions for surface properties here, so that's probably the best place for it for now.

LordOfXen:
All of the tests are complete, and everything works as expected so far.

slouken:
Looks good, thanks!

LordOfXen:
> Looks good, thanks!

My pleasure.

PR #605, opened 2025-08-12
Cumulative fixes for WEBP, AVIF, APNG and GIF Including Timing

LordOfXen:
This PR contains fixes for GIF decoding as well as a more robust implementation for timing, converting `PTS` to an internal-use-only field, and exposing frame-based delay instead, which still relies on timebase_numerator and timebase_denominator.

Here are the changes made:
- Remove `limits.h` from `IMG_avif.c` because it was not being used.
- Remove the `last_pts` field in `IMG_anim_encoder`.
- Remove the `first_pts` field in `IMG_anim_encoder`.
- Introduce the `accumulated_pts` field in `IMG_anim_encoder`.
- Introduce the `last_delay` field in `IMG_anim_encoder`.
- Remove the `GetStreamPresentationTimestampMS` function in `IMG_anim_encoder` because it is no longer needed.
- Introduce the `IMG_GetResolvedDuration` function in `IMG_anim_encoder`.
- Introduce the IMG_GetCurrentTimestamp function in `IMG_anim_encoder`.
- Introduce the `accumulated_pts` field in `IMG_anim_decoder`.
- Introduce the `IMG_CalculateDelay` function in `IMG_anim_decoder`.
- Rename all `pts` parameter names to `delay` in both `SDL_image.h` and all other headers/c files for formats and encoders/decoders.
- Change `IMG_gif`, `IMG_webp`, `IMG_avif`, and `IMG_libpng` to use these new delay calculation functions instead of the old, incorrect algorithms.
- Fix the `IMG_AnimationDecoderGetGIFHeader` function in `IMG_gif` where the extension reader wasn't setting back the original position of the stream before handing back to the parser.

This seems to be implemented more robustly than before, resulting in better consistency for our API as well.

### Tests
The tests continues, please do not merge this until you see the line here change with a different message.

### DIFF
--- a/src/IMG_anim_decoder.c
+++ b/src/IMG_anim_decoder.c
@@ -325,7 +325,7 @@ IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int 
     }
     for (int i = 0; i < anim->count; ++i) {
         anim->frames[i] = frames[i];
-        anim->delays[i] = SDL_round((int)delays[i] * 1000 * decoder->timebase_numerator / decoder->timebase_denominator);
+        anim->delays[i] = SDL_round((int)delays[i] * 1000 * decoder->timebase_numerator / decoder->timebase_denominator); // We do not have decoder here as it's freed above, but that's fine, I think?
     }
 
     SDL_free(frames);
--- a/src/IMG_anim_decoder.c
+++ b/src/IMG_anim_decoder.c
@@ -178,10 +178,10 @@ SDL_PropertiesID IMG_GetAnimationDecoderProperties(IMG_AnimationDecoder *decoder
     return decoder->props;
 }
 
-bool IMG_GetAnimationDecoderFrame(IMG_AnimationDecoder *decoder, SDL_Surface **frame, Uint64 *pts)
+bool IMG_GetAnimationDecoderFrame(IMG_AnimationDecoder *decoder, SDL_Surface **frame, Uint64 *delay)
 {
     SDL_Surface *temp_frame = NULL;
-    Uint64 temp_pts;
+    Uint64 temp_delay;
 
     if (!decoder) {
         return SDL_InvalidParamError("decoder");
@@ -190,14 +190,17 @@ bool IMG_GetAnimationDecoderFrame(IMG_AnimationDecoder *decoder, SDL_Surface **f
     if (!frame) {
         frame = &temp_frame;
     }
-    if (!pts) {
-        pts = &temp_pts;
+    if (!delay) {
+        delay = &temp_delay;
     }
 
-    bool result = decoder->GetNextFrame(decoder, frame, pts);
+    bool result = decoder->GetNextFrame(decoder, frame, delay);
     if (temp_frame) {
         SDL_DestroySurface(temp_frame);
     }
+
+    decoder->accumulated_pts += *delay;
+
     return result;
 }
 
@@ -229,6 +232,14 @@ bool IMG_ResetAnimationDecoder(IMG_AnimationDecoder *decoder)
     }
 
     return decoder->Reset(decoder);
+}
+
+Uint64 IMG_CalculateDelay(IMG_AnimationDecoder* decoder, int delay_num, int delay_den)
+{
+    if (delay_den < 1)
+        delay_den = 100;
+
+    return (Uint64)SDL_round(((double)delay_num / delay_den) * ((double)decoder->timebase_denominator / (double)decoder->timebase_numerator));
 }
 
 IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int maxFrames)
@@ -314,16 +325,7 @@ IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int 
     }
     for (int i = 0; i < anim->count; ++i) {
         anim->frames[i] = frames[i];
-
-        if (i < anim->count - 1) {
-            anim->delays[i] = (int)(delays[i + 1] - delays[i]);
-        } else if (i > 0) {
-            // Assuming consistent frame timing
-            anim->delays[i] = anim->delays[i - 1];
-        } else {
-            // Assuming single frame of 60 FPS animation
-            anim->delays[i] = 16;
-        }
+        anim->delays[i] = SDL_round((int)delays[i] * 1000 * decoder->timebase_numerator / decoder->timebase_denominator);
     }
 
     SDL_free(frames);
--- a/src/IMG_anim_decoder.h
+++ b/src/IMG_anim_decoder.h
@@ -29,12 +29,14 @@ struct IMG_AnimationDecoder
     bool closeio;
     int timebase_numerator;
     int timebase_denominator;
+    Uint64 accumulated_pts;
 
-    bool (*GetNextFrame)(IMG_AnimationDecoder *decoder, SDL_Surface** frame, Uint64* pts);
+    bool (*GetNextFrame)(IMG_AnimationDecoder *decoder, SDL_Surface** frame, Uint64* delay);
     bool (*Reset)(IMG_AnimationDecoder *decoder);
     bool (*Close)(IMG_AnimationDecoder *decoder);
 
     IMG_AnimationDecoderContext *ctx;
 };
 
+extern Uint64 IMG_CalculateDelay(IMG_AnimationDecoder *decoder, int delay_num, int delay_den);
 extern IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int maxFrames);
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -534,7 +534,7 @@ struct IMG_AnimationDecoderContext
     int last_disposal;           /* Disposal method from previous frame */
     int restore_frame;           /* Frame to restore when using DISPOSE_PREVIOUS */
 
-    Uint64 last_pts;
+    //Uint64 last_pts;
 
     char *comment;
     int loop_count;
@@ -618,6 +618,10 @@ static bool IMG_AnimationDecoderGetGIFHeader(IMG_AnimationDecoder *decoder)
             ctx->state.GifScreen.GrayScale = ctx->global_grayscale;
         }
 
+        Uint64 stream_pos = SDL_TellIO(src);
+
+        // TODO: This is an optional part for finding extensions to assign loop count, comment etc.
+        // So maybe all the return SDL_SetError() calls should be replaced with quitting the loop and continuing the execution?
         bool processing_extensions = true;
         while (processing_extensions) {
             uint8_t block_type;
@@ -704,7 +708,6 @@ static bool IMG_AnimationDecoderGetGIFHeader(IMG_AnimationDecoder *decoder)
                 }
             } break;
             case 0x2C: // Image Descriptor
-                SDL_SeekIO(src, -1, SDL_IO_SEEK_CUR);
                 processing_extensions = false;
                 break;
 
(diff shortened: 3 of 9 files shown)


src/IMG_anim_decoder.c
```diff
@@ -314,16 +325,7 @@ IMG_Animation *IMG_DecodeAsAnimation(SDL_IOStream *src, const char *format, int
...
-        if (i < anim->count - 1) {
-            anim->delays[i] = (int)(delays[i + 1] - delays[i]);
-        } else if (i > 0) {
-            // Assuming consistent frame timing
-            anim->delays[i] = anim->delays[i - 1];
-        } else {
-            // Assuming single frame of 60 FPS animation
-            anim->delays[i] = 16;
-        }
+        anim->delays[i] = SDL_round((int)delays[i] * 1000 * decoder->timebase_numerator / decoder->timebase_denominator); // We do not have decoder here as it's freed above, but that's fine, I think?
```
slouken: We are creating our own decoder here, with the default time base of milliseconds, so we can take the delays directly:
```suggestion
        anim->delays[i] = (int)delays[i];
```

src/IMG_anim_decoder.h
```diff
@@ -29,12 +29,14 @@ struct IMG_AnimationDecoder
     bool closeio;
     int timebase_numerator;
     int timebase_denominator;
+    Uint64 accumulated_pts;
 
-    bool (*GetNextFrame)(IMG_AnimationDecoder *decoder, SDL_Surface** frame, Uint64* pts);
+    bool (*GetNextFrame)(IMG_AnimationDecoder *decoder, SDL_Surface** frame, Uint64* delay);
```
slouken:
```suggestion
    bool (*GetNextFrame)(IMG_AnimationDecoder *decoder, SDL_Surface **frame, Uint64 *delay);
```

src/IMG_gif.c
```diff
@@ -618,6 +618,10 @@ static bool IMG_AnimationDecoderGetGIFHeader(IMG_AnimationDecoder *decoder)
             ctx->state.GifScreen.GrayScale = ctx->global_grayscale;
         }
 
+        Uint64 stream_pos = SDL_TellIO(src);
+
+        // TODO: This is an optional part for finding extensions to assign loop count, comment etc.
+        // So maybe all the return SDL_SetError() calls should be replaced with quitting the loop and continuing the execution?
```
slouken: Yes, that seems reasonable. No need to fail here if we can't parse metadata.

src/IMG_libpng.c
```diff
@@ -922,7 +922,7 @@ static SDL_Surface *decompress_png_frame_data(DecompressionContext* context, png
     if (data_size < 0) {
         goto error;
     }
-    if ((Uint64)data_size >= SIZE_MAX) {
+    if ((Uint64)data_size >= SDL_MAX_SINT32) {
```
slouken: size_t is often 64-bit. This isn't equivalent.

sezero: Yet the old check is broken which would always yield to false

The new one should be OK, but the Uint64 cast should be removed as well

sezero: And the error message is stupid, too. I suggest changing it to `SDL_SetError("data size >= INT32_MAX");`

src/IMG_anim_decoder.c
LordOfXen: You are right, applied, thanks!

src/IMG_libpng.c
LordOfXen: That was supposed to be SDL_SIZE_MAX, my bad for replacing it with the incorrect equivalent.

LordOfXen: Fixed!

src/IMG_gif.c
LordOfXen: Question, though, if we neither fail nor log anything, the user won't be able to find anything regarding why it fails, so maybe we should log there?

src/IMG_libpng.c
```diff
@@ -922,8 +922,8 @@ static SDL_Surface *decompress_png_frame_data(DecompressionContext* context, png
     if (data_size < 0) {
         goto error;
     }
-    if ((Uint64)data_size >= SDL_MAX_SINT32) {
-        SDL_SetError("data size >= sizeof(size_t)");
+    if ((Uint64)data_size >= SDL_SIZE_MAX) {
```
sezero: I guess this ckeck makes sense only for 32 bit targets? Otherwise the check always results in false.

LordOfXen: Well, yes, since it's quite impossible to get a file in the size of a 64-bit integer, but a sanity check is a sanity check, though.

sezero: _Say_ that you did, but the check in memory still will fail: A negative check for the Sint64 is already present above, so the cast-to-Uint64 value can not ever be that big.

LordOfXen: Well that's right, it can be removed in this case completely.

LordOfXen: Fixed.

src/IMG_gif.c
LordOfXen: Now the only thing left to discuss is this as mentioned earlier, should we convert the SDL_SetError's to a SDL_Log? If we don't do anything, any failure will be silently ignored, and user won't be able to know why it fails, nor will they be able to provide any input while asking for help.

LordOfXen: My recommendation would be to use this, maybe:
`bool ignoreProps = SDL_GetBooleanProperty(props, IMG_PROP_METADATA_IGNORE_PROPS_BOOLEAN, false);`
Unless the user wants to ignore metadata, we can continue to set an error and fail. If they don't care about the metadata, then we wouldn't process the metadata part at all.

src/IMG_libpng.c
sezero: Maybe check against UINT32_MAX - but whatever

LordOfXen: I think we should bring back SDL_MAX_SINT32, 2 gb limit for a single PNG frame is more than sufficient already

LordOfXen: Or even smaller maybe

sezero: > I think we should bring back SDL_MAX_SINT32, 2 gb limit for a single PNG frame is more than sufficient already

I agree. (That was my original suggestion.)

LordOfXen: Done against SDL_MAX_SINT32

src/IMG_gif.c
LordOfXen: Solved this. Please read my last comment.

LordOfXen:
All of the issues mentioned have been addressed.
- GIF metadata parser now also respects `IMG_PROP_METADATA_IGNORE_PROPS_BOOLEAN` instead of always processing it. Therefore, unless `IMG_PROP_METADATA_IGNORE_PROPS_BOOLEAN` exists, we continue to set SDL error because otherwise we'd have to log, silently ignoring the failure will be hard to debug on the application level.
- Commented-out codes, like those left from the previous implementation of `last_pts`, have been removed from GIF, WEBP, and PNG (AVIF didn't have this).
- `SDL_MAX_SINT32` for data_size in PNG decompression is validated because a single PNG picture shouldn't possibly exceed 2 GB. However, it can be increased to `SDL_MAX_UINT32` later if necessary/needed.

I do not see an issue with those implementations; however, if we do not have any more issues with this PR, we can merge this and then re-open additional PRs if needed to enhance those above.

FYI, tests are complete as well, working as expected.

LordOfXen:
Slight note, I have noticed a few issues with WEBP and AVIF, let me fix those.

LordOfXen:
All of them are fixed now. Here's a list for the key changes:
- All variable/parameter/field names are updated to use `duration` instead of `delay` or `pts`.
- Previously `IMG_GetAnimationDecoderFrame` were returning true even when there weren't any more frames left, this was complicating the usage to be dependent on the frame's validness rather than the return value of the function. Now it returns `false` when there are no more frames either, following what `IMG_GetAnimationDecoderFrame` believed the underlying `GetNextFrame` was doing. If we did not make this change, `IMG_GetAnimationDecoderFrame`'s NULL `temp_frame` implementation would fail with deadlock since the function will always return true unless there is an error (not including no more frames left).
- Extended `IMG_GetCurrentTimestamp` to take duration directly rather than automatically rellying on `last_duration`. This was restrictive in case you wanted to push your own duration.
- Updated WEBP implementation for encoder to use this new `IMG_GetCurrentTimestamp` both for the last timestamp and for each frame, now it results in correct frame compositions with correct timestamps for each frame.
- Updated AVIF decoder to use `durationInTimescales` instead of `ptsInTimescales` otherwise it was getting calculated incorrectly.

Now all works as expected, clarity have been enhanced (previously some parts were stating delay while some parts were using the duration keyword, now all uses duration).

LordOfXen:
@sezero can you re-run the checks? Linux static failed to connect to a Google link I think.

slouken:
I thought getanimation frame would return true and yield a NULL frame at the end of the animation? If we return false we should set an error describing why, but maybe there's a better way to signal the end of animation?

LordOfXen:
I also thought that first but imagining that the return value of bool for a function should usually state the overall status of the function. The actual job of that function is to gather a frame/surface, if we assign a NULL frame, that means it failed to gather one, so why return true?

This made me think that we should return false for clarity when there are no more frames.

slouken:
> I also thought that first but imagining that the return value of bool for a function should usually state the overall status of the function. The actual job of that function is to gather a frame/surface, if we assign a NULL frame, that means it failed to gather one, so why return true?
> 
> This made me think that we should return false for clarity when there are no more frames.

Yes, but the documentation says it returns false on error and to check SDL_GetError(). Mmmm, maybe we should explicitly call SDL_ClearError() and document that the error will be the empty string in that case?

LordOfXen:
Yeah sure but that still kinda feels unprofessional though, but it would work.

We can also return enum instead of bool maybe? I am not sure.

LordOfXen:
Something like this:

```
IMG_DECODER_STATUS_OK = 1
IMG_DECODER_STATUS_GOTFRAME = 2
IMG_DECODER_STATUS_FAILED = 4

IMG_DECODER_STATUS_VALID = IMG_DECODER_STATUS_OK | IMG_DECODER_STATUS_GOTFRAME
```

To use like this:
```c
SDL_Surface* frame;
Uint64 duration;
while ((GetFrame(&frame, &duration) & IMG_DECODER_STATUS_VALID) == IMG_DECODER_STATUS_VALID)
{
// Will execute here with valid frames unless there are errors or no more frame left.
}
```

Note that we can set failed to 0 as well since not setting any other value would indicate failure already.

slouken:
> Something like this:
> 
> ```
> IMG_DECODER_STATUS_OK = 1
> IMG_DECODER_STATUS_GOTFRAME = 2
> IMG_DECODER_STATUS_FAILED = 4
> 
> IMG_DECODER_STATUS_VALID = IMG_DECODER_STATUS_OK | IMG_DECODER_STATUS_GOTFRAME
> ```
> 
> To use like this:
> 
> ```c
> SDL_Surface* frame;
> Uint64 duration;
> while ((GetFrame(&frame, &duration) & IMG_DECODER_STATUS_VALID) == IMG_DECODER_STATUS_VALID)
> {
> // Will execute here with valid frames unless there are errors or no more frame left.
> }
> ```
> 
> Note that we can set failed to 0 as well since not setting any other value would indicate failure already.

Let's call SDL_ClearError() for now and come back to this as a separate topic.

LordOfXen:
Okay will do so in a bit.

LordOfXen:
@slouken Updated the comment regarding this and added SDL_ClearError before the call to GetNextFrame as you said.

When this gets merged, I will make an issue regarding this topic for discussion.

slouken:
Merged!

Issue #607, opened 2025-08-13
Build fails on iOS


smoogipoo:
Hello! The build as of https://github.com/libsdl-org/SDL_image/commit/cc178bea511aa22258ab6c37650826b564f54ad7 appears to fail on iOS targets.

```
[...]

PR #609, opened 2025-08-13
Add missing compilation files to Xcode project

smoogipoo:
Fixes https://github.com/libsdl-org/SDL_image/issues/607

Added/changed in https://github.com/libsdl-org/SDL_image/pull/600, the files were not present in the Xcode project causing iOS builds to fail.

Passing run: https://github.com/smoogipoo/SDL3-CS/actions/runs/16942815791


slouken:
Thanks, this is fixed in https://github.com/libsdl-org/SDL_image/commit/90e8ee775595fee870325fc11b8e0291d84a8a6b

PR #611, opened 2025-08-14
IMG_CODER_STATUS Feature For En/De-coders

LordOfXen:
Introducing `IMG_CODER_STATUS`, which is used in symmetry by both encoders and decoders for animation.

Here's a brief list:
- Added `IMG_CODER_STATUS` enum.
- Added `IMG_CODER_STATUS_OK` for stating that the coder was successful.
- Added `IMG_CODER_STATUS_FAILED` for stating that the coder was unsuccessful.
- Added `IMG_CODER_STATUS_MAX` for stating that we either hit max available frames (for decoders) or that we are not allowed to add more frames (for encoders).
- Added `IMG_CODER_STATUS_INVALID` for stating when `IMG_GetAnimationEncoderStatus` or `IMG_GetAnimationDecoderStatus` is called with an invalid coder parameter. It can be used in other similar cases as well.
- Modified `IMG_AddAnimationEncoderFrame` and `IMG_GetAnimationDecoderFrame` to first assign `IMG_CODER_STATUS_OK` and then, after the call to the underlying function, check if the status is still OK, but if an error has been provided by the underlying coder, we set `IMG_CODER_STATUS_FAILED` automatically. This smoothly handles statuses at the coder level without having to interfere with the existing `SDL_SetError` calls. However, calls for `IMG_CODER_STATUS_MAX` and such are still provided by the underlying coders.
- Modified WEBP, AVIF, PNG, and GIF accordingly to use this new `IMG_CODER_STATUS` (e.g., `IMG_CODER_STATUS_MAX`).
- Small other additions where brackets were forgotten, added those.

This does not interfere with the execution and can safely be merged in after an approval for the design.

### DIFF
--- a/src/SDL_image.sym
+++ b/src/SDL_image.sym
@@ -83,5 +83,6 @@ SDL3_image_0.0.0 {
     IMG_ResetAnimationDecoder;
     IMG_CloseAnimationDecoder;
     IMG_GetAnimationDecoderProperties;
+    IMG_GetAnimationDecoderStatus;
   local: *;
 };


LordOfXen:
Maybe we can rename `IMG_CoderStatus` to `IMG_AnimationCoderStatus`, up to you guys.

include/SDL3_image/SDL_image.h
```diff
@@ -2403,6 +2403,18 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
 extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadWEBPAnimation_IO(SDL_IOStream *src);
 
+/**
+* An enum representing the status of the encoder and decoder.
+*/
+typedef enum IMG_CoderStatus
+{
+	IMG_CODER_STATUS_OK, 	    /**< En/De-coded the frame successfully. */
+	IMG_CODER_STATUS_FAILED,    /**< En/De-coding the frame failed. Call SDL_GetError for more information. */
+	IMG_CODER_STATUS_MAX, 	    /**< En/De-coding reached maximum frames. For Decoding, this means no more frames available. */
```
```suggestion
	IMG_CODER_STATUS_COMPLETE, 	    /**< The animation is complete and no more frames are available. */
```

include/SDL3_image/SDL_image.h
```diff
@@ -2403,6 +2403,18 @@ extern SDL_DECLSPEC IMG_Animation * SDLCALL IMG_LoadGIFAnimation_IO(SDL_IOStream
...
+
+    IMG_CODER_STATUS_INVALID,   /**< Invalid coder status that does not represent a decoder or encoder status. */
+} IMG_CoderStatus;
```
```suggestion
} IMG_DecodeStatus;
```

include/SDL3_image/SDL_image.h
```diff
@@ -2519,7 +2531,21 @@ extern SDL_DECLSPEC IMG_AnimationEncoder * SDLCALL IMG_CreateAnimationEncoderWit
...
+*
+* \param encoder the encoder to get the status of.
+* \returns the status of the underlying encoder, or IMG_CODER_STATUS_INVALID if the given
+*          encoder is invalid.
+*
+* \since This function is available since SDL_image 3.4.0.
+*
+* \sa IMG_AddAnimationEncoderFrame
+*/
+extern SDL_DECLSPEC IMG_CoderStatus SDLCALL IMG_GetAnimationEncoderStatus(IMG_AnimationEncoder *encoder);
```
slouken: We don't need this function for symmetry, we only need to disambiguate between error and EOF in the decode case.

LordOfXen:
Okay if we only need this for the decoder part I'll update it all soon to only reflect this change for our decoder.
Should I rename IMG_CoderStatus to IMG_AnimationDecoderStatus?

src/IMG_anim_decoder.c
```diff
@@ -193,14 +193,21 @@ bool IMG_GetAnimationDecoderFrame(IMG_AnimationDecoder *decoder, SDL_Surface **f
...
 
     bool result = decoder->GetNextFrame(decoder, frame, duration);
     if (temp_frame) {
         SDL_DestroySurface(temp_frame);
     }
 
+    // If the underlying decoder only uses the SDL_SetError function, we can update our status accordingly as well.
+    if (decoder->status == IMG_CODER_STATUS_OK && SDL_GetError()[0] != '\0') {
+        decoder->status = IMG_CODER_STATUS_FAILED;
+    }
```
slouken:
```suggestion
    // If the underlying decoder only uses the SDL_SetError function, we can update our status accordingly as well.
    if (!result) {
        if (SDL_GetError()[0] == '\0') {
            decoder->status = IMG_CODER_STATUS_COMPLETE;
        } else {
            decoder->status = IMG_CODER_STATUS_FAILED;
        }
    }
```

LordOfXen: This yields to the same thing doesn't it? We clear the error before the call to the function, if any error is set then the return value doesn't matter I believe. (Just making sure you are aware)

src/IMG_avif.c
```diff
@@ -890,13 +890,15 @@ static bool IMG_AnimationDecoderGetNextFrame_Internal(IMG_AnimationDecoder *deco
     avifResult result;
 
     if (ctx->total_frames - ctx->current_frame < 1) {
+        decoder->status = IMG_CODER_STATUS_MAX;
```
slouken: These aren't necessary with the change I made above.

slouken:
> Okay if we only need this for the decoder part I'll update it all soon to only reflect this change for our decoder. Should I rename IMG_CoderStatus to IMG_AnimationDecoderStatus?

I suggested IMG_DecodeStatus above. Could be IMG_AnimationDecodeStatus?

src/IMG_anim_decoder.c
slouken: SDL functions can set the error internally and then recover successfully, leaving the error still set.

LordOfXen: Ah I get what you did, so we don't rely on the decoder parts at all, got it.

LordOfXen:
Alright IMG_AnimationDecodeStatus it is then, will make those soon.

LordOfXen:
```c
typedef enum IMG_CoderStatus
{
	IMG_DECODE_STATUS_OK, 	            /**< Decoded the frame successfully. */
	IMG_DECODE_STATUS_FAILED,           /**< Decoding the frame failed. Call SDL_GetError for more information. */
	IMG_DECODE_STATUS_COMPLETE, 	    /**< No more frames available. */

    IMG_DECODE_STATUS_INVALID,          /**< Invalid decoder status that does not represent any valid status. */
} IMG_CoderStatus;
```

@slouken should we keep IMG_DECODE_STATUS_INVALID? if we won't, what will it be when the user gives NULL to decoder pointer while calling the GetAnimationDecoderStatus?

Do you want function to be still named GetAnimationDecoderStatus? If not, what do you want it to be named?

Please let me know these.

(Ignore the enum name, haven't changed that yet, just check the fields)


FYI: IMG_AnimationDecodeStatus sounds grammatically incorrect; should it be IMG_AnimationDecoderStatus or IMG_AnimationFrameDecodeStatus? Or do you want to keep DecodeStatus as-is?

slouken:
> @slouken should we keep IMG_DECODE_STATUS_INVALID? if we won't, what will it be when the user gives NULL to decoder pointer while calling the GetAnimationDecoderStatus?

Yes, keeping it seems correct.

> Do you want function to be still named GetAnimationDecoderStatus? If not, what do you want it to be named?

Sure, that sounds good.

> FYI: IMG_AnimationDecodeStatus sounds grammatically incorrect; should it be IMG_AnimationDecoderStatus or IMG_AnimationFrameDecodeStatus? Or do you want to keep DecodeStatus as-is?

It sounds grammatically correct to me, but IMG_AnimationDecoderStatus is probably better. It's the current status of the decoder.

LordOfXen:
Done, also fixed a bug in the encoder part where we weren't checking the result before increasing the accumulated_pts and last_delay with the given duration, it was getting increased even when the function failed, now that won't be the case.

src/IMG_anim_decoder.c
```diff
@@ -193,14 +193,24 @@ bool IMG_GetAnimationDecoderFrame(IMG_AnimationDecoder *decoder, SDL_Surface **f
     if (!duration) {
         duration = &temp_duration;
     }
-
+
```
slouken:
```suggestion

```

src/IMG_anim_decoder.h
```diff
@@ -23,6 +23,8 @@ typedef struct IMG_AnimationDecoderContext IMG_AnimationDecoderContext;
 
 struct IMG_AnimationDecoder
 {
+    IMG_AnimationDecoderStatus status;
+
```
slouken:
```suggestion
```

src/IMG_anim_encoder.c
```diff
@@ -171,11 +171,13 @@ bool IMG_AddAnimationEncoderFrame(IMG_AnimationEncoder *encoder, SDL_Surface *su
     if (!surface || surface->w <= 0 || surface->h <= 0) {
         return SDL_InvalidParamError("surface");
     }
-
+
```
slouken:
```suggestion

```

include/SDL3_image/SDL_image.h
```diff
@@ -2519,7 +2531,7 @@ extern SDL_DECLSPEC IMG_AnimationEncoder * SDLCALL IMG_CreateAnimationEncoderWit
  * \sa IMG_CreateAnimationEncoderWithProperties
  * \sa IMG_CloseAnimationEncoder
  */
-extern SDL_DECLSPEC bool SDLCALL IMG_AddAnimationEncoderFrame(IMG_AnimationEncoder *encoder, SDL_Surface *surface, Uint64 duration);
+extern SDL_DECLSPEC bool SDLCALL IMG_AddAnimationEncoderFrame(IMG_AnimationEncoder *encoder, SDL_Surface *surface, Uint64 delay);
```
slouken: Was this change intentional? I think it's duration everywhere, right?

include/SDL3_image/SDL_image.h
```diff
@@ -2688,6 +2700,19 @@ extern SDL_DECLSPEC SDL_PropertiesID SDLCALL IMG_GetAnimationDecoderProperties(I
...
+*
+* \param decoder the decoder to get the status of.
+* \returns the status of the underlying decoder, or IMG_DECODER_STATUS_INVALID if the given
+*          decoder is invalid.
+*
+* \since This function is available since SDL_image 3.4.0.
+*
+* \sa IMG_GetAnimationDecoderFrame
+*/
+extern SDL_DECLSPEC IMG_AnimationDecoderStatus SDLCALL IMG_GetAnimationDecoderStatus(IMG_AnimationDecoder *decoder);
```
slouken, line 2714: This needs to be added to SDL_image.sym

include/SDL3_image/SDL_image.h
LordOfXen: Nah it wasn't intentional, my header is probably behind the official one. I'll just update it with the latest after this pr.

I will revert that to duration as well, thanks for catching!

include/SDL3_image/SDL_image.h
LordOfXen: Done!

slouken:
Merged, thanks!

LordOfXen:
> Merged, thanks!

Thanks to you as well.

PR #615, opened 2025-08-16
miniz: Avoid arithmetic on null pointer to avoid UBSan

jayschwa:
Clang/LLVM's undefined behavior sanitizer triggers a crash when an offset (even zero) is added to a null pointer. I used a debugger to confirm that `num_bytes_to_process` is zero, which means there should be no change in behavior.

The next revision of the C language standard (C2Y) allows zero-length operations on null pointers, but we may be waiting a couple years before it is finalized.

See: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf


sezero:
This is a backport of https://github.com/richgel999/miniz/pull/227, yes?

Merging. Will change the last pSrc to NULL for simplicity and better symmetry with mainstream.

jayschwa:
It looks like the main (SDL3) branch will also need this patch.

sezero:
> It looks like the main (SDL3) branch will also need this patch.

Cherry-picked everywhere.

PR #618, opened 2025-08-25
SDL_image cleanup

slouken:


slouken:
@LordOfXen, can you please review?

LordOfXen:
Seems OK, tests are OK as well.
Though we may need further cleanup, for example __xmlman_ConstructXMPWithRDFDescription outputs len in the last parameter, which is redundant because we always return a null-terminated string, so SDL_strlen would always do instead of that.

Not considering an additional functional call over a parameter overhead in a function is an option that you can decide.

slouken:
Okay, merged. Feel free to do further cleanup, especially around the XML parsing and the FIXME I added.

PR #622, opened 2025-09-02
XMLMan Fixes & Test Suite Improvements

LordOfXen:
Implements a new XMP parser for XML reading:
- Reduced allocations to 1; the unescape and other functions directly rely on the allocated memory by moving things around. We move the null-terminator accordingly while applying the unescaping character, leaving the garbage operators behind the terminator; this is less costly than allocating memory separately. This directly fixes the FIXME left by @slouken.
- Entirely rewritten parsing logic, no longer relying on null-terminators, rather we rely on the given len as supposed to.
- Ignores XML comments and CDATA; this is an addition for edge cases since comments or CDATA are less likely to be encountered in XMP data.
- Case-insensitive search for "en-US" but case-sensitive search for XML tags as expected by the spec.
- Added third-party encoded XMP metadata test to be decoded by SDL and get the expected string with the correct size/len.

@sezero Please check your valgrind as we did before to make sure the behaviour is the same as before.


sezero:
> @sezero Please check your valgrind as we did before to make sure the behaviour is the same as before.

valgrind emitted no errors for me with animation testsuite on i686-linux.

```
==16435== HEAP SUMMARY:
==16435==     in use at exit: 53,286,968 bytes in 1,969 blocks
==16435==   total heap usage: 33,496 allocs, 31,527 frees, 354,365,410 bytes allocated
==16435== 
==16435== LEAK SUMMARY:
==16435==    definitely lost: 40,856 bytes in 200 blocks
==16435==    indirectly lost: 1,068,016 bytes in 84 blocks
==16435==      possibly lost: 52,080,666 bytes in 264 blocks
==16435==    still reachable: 97,430 bytes in 1,421 blocks
==16435==         suppressed: 0 bytes in 0 blocks
==16435== Rerun with --leak-check=full to see details of leaked memory
==16435== 
==16435== For counts of detected and suppressed errors, rerun with: -v
==16435== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 74 from 13)
```

LordOfXen:
> valgrind emitted no errors for me with animation testsuite on i686-linux.

Enabled WEBP and AVIF as before? Just asking for clarity since previously it wasn't emitting errors either without enabling either of those (I think it was AVIF).

sezero:
> Enabled WEBP and AVIF as before?

Yes

LordOfXen:
> Yes

Alright then all good. Thank you.

slouken:
Merged, thanks!

LordOfXen:
My pleasure.

PR #623, opened 2025-09-03
cmake: build and run animation tests

### LINKED ISSUE #621: Animation Test Suite Not Enforced
The animation test suite written for all-case scenarios is not run by CI for the upcoming pull requests.
We should add it to CI as soon as possible to enforce fixed behaviour.

madebr:
The animation tests are now built by CMake, and run with CTest.
I've also slightly rewritten the file to use SDL_test's test suite.

The tests currently fail because by default, SDL_image is built without libpng support (or at least can be built without png/avif/... support).
The other test suite (testimage), uses environment variables that check whether a certain test should be disabled.
e.g. it disables jpeg tests if the `SDL_IMAGE_SAVE_JPG` environment variable is set to `0`.

Fixes #621


sezero:
What is the change to vendored aom about? Seems to go back and revert the mingw32 fix, therefore seems wrong to me.

madebr:
> What is the change to vendored aom about? Seems to go back and revert the mingw32 fix, therefore seems wrong to me.

Yup, accidental push. I had non-committed changes so `git submodule update` failed.

slouken:
Can we use `#ifdef LOAD_PNG` (etc.) to gate various formats in the animation tests?

madebr:
> Can we use `#ifdef LOAD_PNG` (etc.) to gate various formats in the animation tests?

Yes, but smcv used environment variables such that the test binary is independent of a build configuration:
you have more confidence in your test suite if it is more or less decoupled from the thing it is testing.

https://github.com/libsdl-org/SDL_image/pull/258#discussion_r880346699

LordOfXen:
Assuming this gets to run regardless of the CMake flag preferences, it looks OK to me.

LordOfXen:
This code was initially taken from my C++ test unit while converting it to C. It was in fact reporting the result of tests like: PASSES or FAIL. Here (not related to @madebr's changes), it serves no purpose but as a garbage log pile.
```C

    SDLTest_Log("--TESTS--");

    for (size_t cim = 0; cim < SDL_arraysize(inputImages); ++cim) {

        for (size_t com = 0; com < SDL_arraysize(outputImageFormats); ++com) {

            const char *inputImage = inputImages[cim].filename;

            const char *outputImageFormat = outputImageFormats[com];

            SDLTest_Log("%s >> %s == %s", inputImage, outputImageFormat, inputImage);

            if (com == 0) {

                SDLTest_Log("--- Rewind Tests for %s ---", inputImage);

                SDLTest_Log("--- Metadata Tests for %s ---", inputImage);

            }

        }

    }
```

Unless we have a plan to extend this to list the results in the end (which requires the entire test to be changed as then we should not early return), I believe we can also get rid of it.
We already log every single test one by one where required with possible information in them.

slouken:
Merged, thanks!

PR #628, opened 2025-09-22
Enhance xmlman

### LINKED ISSUE #627: Buffer Overflow/Heap Corruption in xml_unescape_inplace when input fills buffer exactly.
https://github.com/libsdl-org/SDL_image/blob/eeae8a64dfc065918fc259a7afe271bce8bcaadc/src/xmlman.c#L75

A heap buffer overflow occurs in `xml_unescape_inplace` when the input string has no XML entities and its length equals the buffer size (`len`). The function writes a null terminator at `str[j]` without checking if `j < len`, causing a 1-byte out-of-bounds write.

This can be triggered with a simple input like `"abcde"` and `len = 5`.
or
char buf[4] = "&&&&"; // 4 bytes: all '&'
xml_unescape_inplace(buf, 4);

✅ Fix: Add bounds check before null termination:
```c
if (j < len) {
    str[j] = '\0';
} else if (len > 0) {
    str[len - 1] = '\0';
}

OR

while (i < len && j < len && str[i] != '\0') {  // 👈 加上 j < len
    ...
}

### PR DESCRIPTION
**Pull request by LordOfXen**
Created to fix #627 and to enhance the overall code of XML Manager.

Here's a brief list:
- Added the validation check to `__xmlman_unescape_inplace` (previously `xml_unescape_inplace`) as recommended by @nopnopnop-lavine.
- Renamed all non-prefixed functions to include the __xmlman prefix (no external change needed, all of those are internal functions).
- Added the header (License text and such).
- Corrected spacing for some lines.

For clarity, the added validation to `__xmlman_unescape_inplace` is this:
```C
if (j < len) {
    str[j] = '\0';
} else if (len > 0) {
    str[len - 1] = '\0';
}
```

This sacrifices correctness over vulnerabilities, as this fix only applies to situations where standalone calls are providing a no-op input for this function, the check then replaces the last character with the null-terminator, rather than writing *outside* the buffer. This results in incorrect output, but without any overflow.

### DIFF
--- a/src/xmlman.c
+++ b/src/xmlman.c
@@ -273,10 +273,11 @@ static char* __xmlman_GetXMLContentFromTag(const char* data, size_t len, const c
             
             if (default_lang || en_us_lang) {
                 size_t content_len = li_end - li_content_start;
-                char* result = (char*)SDL_malloc(content_len + 1);
+                ++content_len;
+                char* result = (char*)SDL_malloc(content_len);
                 if (result) {
-                    SDL_memcpy(result, li_content_start, content_len);
-                    result[content_len] = '\0';
+                    SDL_memcpy(result, li_content_start, content_len - 1);
+                    result[content_len - 1] = '\0';
                     __xmlman_unescape_inplace(result, content_len);
                 }
                 return result;
@@ -303,10 +304,11 @@ static char* __xmlman_GetXMLContentFromTag(const char* data, size_t len, const c
                 return NULL;
             }
 
-            char* result = (char*)SDL_malloc(fallback_len + 1);
+            ++fallback_len;
+            char* result = (char*)SDL_malloc(fallback_len);
             if (result) {
-                SDL_memcpy(result, fallback_content, fallback_len);
-                result[fallback_len] = '\0';
+                SDL_memcpy(result, fallback_content, fallback_len - 1);
+                result[fallback_len - 1] = '\0';
                 __xmlman_unescape_inplace(result, fallback_len);
             }
             return result;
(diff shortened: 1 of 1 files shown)


src/xmlman.c
```diff
@@ -103,11 +126,16 @@ static size_t xml_unescape_inplace(char* str, size_t len) {
...
+        str[j] = '\0';
+    } else if (len > 0) {
+        str[len - 1] = '\0';
+    }
+
     return j;
 }
 
-static const char* find_char_in_bounds(const char* str, const char* end, char c) {
+static const char* __xmlman_find_char_in_bounds(const char* str, const char* end, char c) {
```
slouken: You don't need to namespace or double-underscore static functions. These are file scope only.

LordOfXen: It was odd that some were snake-case, some were not. I'll just make them all snake-case and remove the prefix (for the internal functions).

src/xmlman.c
```diff
@@ -103,7 +126,12 @@ static size_t xml_unescape_inplace(char* str, size_t len) {
         }
     }
 
-    str[j] = '\0';
+    if (j < len) {
+        str[j] = '\0';
+    } else if (len > 0) {
```
sezero, line 131: 1. I think `len > 0` check is not necessary, because you check for `len == 0`at top, yes?

2. If we ever reach here, should we still return `j`, or should we return `len - 1` ?

LordOfXen: @sezero Yes, so the second check can be simplified to a direct else condition.
The top condition can be extended to be `< 1` as well, but we can talk about this later in another PR, maybe? Since it works for now.

Or go ahead and replace `else if` with `else` and modify the top condition to be `len < 1` instead of `len == 0`, making a PR for this kind of small changes feels odd.

sezero: > The top condition can be extended to be `< 1` as well, but we can talk about this later in another PR, maybe?

`len` is `size_t`, i.e. unsigned, therefore `== 0` is good already.

sezero:
@LordOfXen : Is the following good?

```diff
diff --git a/src/xmlman.c b/src/xmlman.c
index db4c4e0..e6498e7 100644
--- a/src/xmlman.c
+++ b/src/xmlman.c
@@ -126,11 +126,10 @@ static size_t unescape_inplace(char* str, size_t len) {
         }
     }
 
-    if (j < len) {
-        str[j] = '\0';
-    } else if (len > 0) {
-        str[len - 1] = '\0';
+    if (j >= len) {
+        j = len - 1;
     }
+    str[j] = '\0';
 
     return j;
 }
```

src/xmlman.c
LordOfXen: Oh right, then yes, that check at the beginning is OK.

LordOfXen:
> @LordOfXen : Is the following good?

Yes, that looks OK to me, thank you.

sezero:
> Yes, that looks OK to me, thank you.

Applied as https://github.com/libsdl-org/SDL_image/commit/d354e3d5146117f8b2f14096800965e56f9f7bfc

PR #637, opened 2025-10-18
Optimize GIF decoder with better transparency handling

### LINKED ISSUE #632: Loading gifsicle optimized transparent GIF fails with "Failed to set transparency on frame"
The original GIF (below) loads fine using `IMG_LoadAnimation_IO`, but after using gifsicle to optimize it with the default/lowest level of optimization, it fails to load. The optimized file loads fine with working transparency in browsers etc. I've attached both GIFs.

Stable SDL_image loads the GIF, but without transparency. The dev version in main produces the error.

The documentation for gifsicle says:
`> -O1  Store only the changed portion of each image. This is the default.`
(I haven't tried higher optimization levels as O1 already fails)

Original: Image
Optimized: Image

LordOfXen:
This patch fixes #632 as well as handling a FIXME in the invalid frame case.

Here's a brief list:
- If we fail to decode a frame of GIF, the entire decoding is set to fail. It feels like a better approach than proceeding with "Success" status and then showing an incomplete animation on the user side.
- Previously, the frames were retrieved in INDEX8 format, then blitted into an RGBA32 canvas. Now we directly create each frame as RGBA32 surfaces with transparency mapped to the appropriate alpha channel of each pixel, resulting in smooth decoding for every image.

Tested with the original rgbrgb.gif, tested with the images given by @maia-s in #632, also tested with 2 different solid and transparent GIF animations, seems to decode all of them correctly so far.

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -891,16 +891,18 @@ static bool IMG_AnimationDecoderGetNextFrame_Internal(IMG_AnimationDecoder *deco
                               BitSet(ctx->buf[8], INTERLACE), 0, &ctx->state);
         }
 
-        if (!image) {
-            // Incorrect animation is harder to detect than a direct failure,
-            // so it's better to fail than try to animate a GIF without a,
-            // full set of frames it has in the file.
+      if (!image) {
+          // Incorrect animation is harder to detect than a direct failure,
+          // so it's better to fail than try to animate a GIF without a,
+          // full set of frames it has in the file.
 
-            // Only set the error if ReadImage did not do it.
-            if (SDL_GetError()[0] == '\0') {
-                return SDL_SetError("Failed to decode frame.");
-            }
-        }
+          // Only set the error if ReadImage did not do it.
+          if (SDL_GetError()[0] == '\0') {
+              return SDL_SetError("Failed to decode frame.");
+          }
+
+          return false;
+      }
 
         /* Composite the frame onto the canvas */
         SDL_Rect dest = { left, top, width, height };


maia-s:
I can confirm that this fixes #632 for me! Thank you!

slouken:
We should keep GIFs as indexed images if possible, to retain previous behavior and reduce memory size. Is it possible to do that here?

LordOfXen:
> We should keep GIFs as indexed images if possible, to retain previous behavior and reduce memory size. Is it possible to do that here?

I do agree, but I literally changed nothing other than moving the temp image from INDEX8 to RGBA32.
The actual canvas, which lives as long as the decoder lives, was always RGBA32 even before this patch; the part I modified only lives until it reaches the blit to move the temp image into the canvas, then it gets destroyed.

So to make this indexed, we must change the canvas itself to be INDEX8, which is not related to this change here.

LordOfXen:
Maybe we can push this for now to fix the bug, until we implement a proper/fully working INDEX8 version for both canvas and the temp image.

slouken:
If this only affects animated GIFs, we can accept this patch as-is.

LordOfXen:
> If this only affects animated GIFs, we can accept this patch as-is.

Not really, single-frame GIF images are decoded using the animation back-end as well.

slouken:
Then I'll accept this patch and we need to figure out how to make indexed GIFs indexed surfaces in the single image decode case.

LordOfXen:
> Then I'll accept this patch and we need to figure out how to make indexed GIFs indexed surfaces in the single image decode case.

When I find time, I'll work on moving it entirely to use INDEX8 over RGBA32, especially the canvas one, since that's more important/is the one held in the background.

maia-s:
Note that each frame in an animated gif can have its own palette, so when composing an image based on multiple gif frames it's possible for the resultant image to have more than 256 colors

slouken:
> Note that each frame in an animated gif can have its own palette, so when composing an image based on multiple gif frames it's possible for the resultant image to have more than 256 colors

For animations it makes sense to have the canvas be RGBA, but loading a single image should preserve the palette.

PR #638, opened 2025-10-20
Added support for animated cursors

slouken:
This is a good test cursor, with author/title metadata, alternate sizes, and PNG images:
https://sweezy-cursors.com/cursor/orange-butterfly-animated


CMakeLists.txt
```diff
@@ -594,6 +597,21 @@ if(SDLIMAGE_AVIF)
     endif()
 endif()
```
madebr, line 598: ani comes before avid alphabetically :)

If you add ANI support for big icons with PNG encoding, then you might have to move it down to make sure `SDLIMAGE_PNG_ENABLED` and `SDLIMAGE_PNG_SAVE` are defined.

slouken: > ani comes before avid alphabetically :)

I didn't want to insert in between the AVIF related options or next to the other non-alphabetized backends. Feel free to suggest a better location for this?

> If you add ANI support for big icons with PNG encoding, then you might have to move it down to make sure `SDLIMAGE_PNG_ENABLED` and `SDLIMAGE_PNG_SAVE` are defined.

I already have, I just call IMG_LoadPNG_IO() in the ICO/CUR code, and it resolves naturally.

PR #650, opened 2025-11-02
Add Android ELF Alignment check

elevenfive:
Adding check to verify that the AAR will meet the [Google Play requirements](https://developer.android.com/guide/practices/page-sizes) already enforced:

```
November 1st, 2025, all new apps and updates to existing apps submitted to Google Play and targeting Android 15+ devices must support 16 KB page sizes on 64-bit devices.
```

@madebr can we get a new release of SDL3_image that will work with the new Play requirements?  The fix was already done a few months ago in d21668694ee468678cc40f3f07230c6349de5d83


build-scripts/build-release.py
```diff
@@ -1398,7 +1398,7 @@ def main(argv=None) -> int:
     parser.add_argument('--vs-year', dest="vs_year", help="Visual Studio year")
     parser.add_argument('--android-api', dest="android_api", help="Android API version")
     parser.add_argument('--android-home', dest="android_home", default=os.environ.get("ANDROID_HOME"), help="Android Home folder")
-    parser.add_argument('--android-ndk-home', dest="android_ndk_home", default=os.environ.get("ANDROID_NDK_HOME"), help="Android NDK Home folder")
+    parser.add_argument('--android-ndk-home', dest="android_ndk_home", default=os.environ.get("ANDROID_NDK_LATEST_HOME"), help="Android NDK Home folder")
```
madebr: Is this a standard Android NDK environment variable?
We want this script to not depend too much on GitHub Actions details.

elevenfive: Honestly no but I'm not sure what else to do.  It is annoying to have this baked-in-to-the-image dependency.  I don't think GH will break this too readily as we would not be the first project to use it.

madebr: The workflow itself is supposed to pass variables.
[This is how it's done for SDL3](https://github.com/libsdl-org/SDL/blob/63cb046d38c9db7598290b5a7b0a8f624dbe4a07/.github/workflows/release.yml#L613-L620). Looks like some arguments are missing in SDL3_image.

elevenfive: LMK what you think, I found and have proposed a minimal fix.  

My thinking is that I would like to avoid a full-scale deep dive into the deeper NDK versioning issues.  Let's file an issue and I can look into that next in a more relaxed timeframe?

Here is the release build dry-run that passes verification: 

https://github.com/elevenfive/SDL_image/actions/runs/19019880714/job/54313296844

elevenfive:
Looks good!  Thanks  both of you!

https://github.com/elevenfive/SDL_image/actions/runs/19022483116/job/54320025489

```
--android-ndk-home "/opt/hostedtoolcache/ndk/r28c/x64" 
```

```
=== ELF alignment ===
/tmp/SDL3_image-android/lib/arm64-v8a/libSDL3_image.so: ALIGNED (2**14)
=====================
ELF Verification Successful
=== ELF alignment ===
/tmp/SDL3_image-android/lib/x86_64/libSDL3_image.so: ALIGNED (2**14)
=====================
ELF Verification Successful
```

elevenfive:
Let's keep this a clean change for just the Android part of things. The Mac/iOS part really doesn't belong in this pr, eh?  I will revert back to minimal Android-only change @madebr .

Release workflow dry-run triggered: https://github.com/elevenfive/SDL_image/actions/runs/19048979040

elevenfive:
I think it's probably best you two handle the underlying build issues holistically (as you mentioned across all the satellite SDL* projects). I'll go ahead and close these PRs and I'll be able to unblock my own project with a local build with the correct NDK version (r28c). Thanks for your help on this!

PR #651, opened 2025-11-04
release: apply updates from SDL3 repo

madebr:
- test Android elf alignment for x86_64/aarch64
- test cmake support of frameworks inside xcframework

The tests in `dmg-verify` **require** a new SDL3 3.2.x/3.4 release as they depend on https://github.com/libsdl-org/SDL/commit/024221d448d8e9b6f50cf1d97249008b55c0e344


madebr:
@slouken
I *think* the dmg-verify step will succeed after you do a new SDL 3.2.x/3.4 release, but make sure to do a test release build before tagging the release.

Issue #645, opened 2025-10-22
Animation encoder transparency is broken


slouken:
The easiest way to reproduce this is to run:
`showanim -save butterfly.gif butterfly.ani`

This example shows GIF encoding, but it looks like it's broken for all formats.

slouken:
@LordOfXen, can you take a look at this?

Here's the butterfly animation I was using for testing:
https://sweezy-cursors.com/cursor/orange-butterfly-animated/

LordOfXen:
Could you tell me what's exactly wrong with formats other than GIF?, I tried the butterfly GIF animation to create a WebP or PNG animation, and no transparency issues occurred.

slouken:
I didn't verify that other formats were broken and I don't remember why I thought they would be affected. If you're not seeing any problems, we can ignore that feedback for now.

LordOfXen:
Oh, alright, they seem to work. I performed the same tests as with GIF, and the results were always correct animation outputs.
(Quick note though, I only tested WebP and APNG, did not do the tests for AVIFS)

PR #654, opened 2025-11-06
Patch for GIF Decoding/Encoding

### LINKED ISSUE #641: GIF images should have a palette
GIF animations should be 32-bit to handle composing multiple frames with different colormaps, but static images should be loaded as INDEX8 surfaces with a palette.

This is a regression and needs to be fixed for the 3.4.0 release.

LordOfXen:
Refactor GIF image processing to support palette management and transparency handling. This fixes #641 and #645.
Note that I could not reproduce the transparency issue with encoding using formats other than GIF.

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -449,7 +449,7 @@ ReadImage(SDL_IOStream * src, int len, int height, int cmapSize,
     }
 
     if (state->Gif89.transparent >= 0 && state->Gif89.transparent < cmapSize) {
-        SDL_SetSurfaceColorKey(image, 1, state->Gif89.transparent);
+        SDL_SetSurfaceColorKey(image, true, state->Gif89.transparent);
     }
 
     while ((v = LWZReadByte(src, FALSE, c, state)) >= 0) {


src/IMG_gif.c
```diff
@@ -423,45 +423,37 @@ ReadImage(SDL_IOStream * src, int len, int height, int cmapSize,
...
-            } else {
-                /* Invalid color index */
-                dst[0] = 255;
-                dst[1] = 0;
-                dst[2] = 255;
-                dst[3] = 255;
-            }
-        }
+    if (state->Gif89.transparent >= 0 && state->Gif89.transparent < cmapSize) {
+        SDL_SetSurfaceColorKey(image, 1, state->Gif89.transparent);
```
slouken:
```suggestion
        SDL_SetSurfaceColorKey(image, true, state->Gif89.transparent);
```

slouken:
Great, did you confirm that both of those issues are fixed?

LordOfXen:
> Great, did you confirm that both of those issues are fixed?

I've run the animation test suite manually, tested with your mentioned test image, butterfly, and also tested with the previous issue that we solved by switching to RGBA, to make sure switching back to INDEX8 did not break it, I also tried to decode and encode a non-transparent GIF to make sure that also stays correct. All of them passed with no issues, and we got the expected results.

Not sure what else to test with, but right now, all those passes.
If you'd like, you can give it a try as well to make sure.

slouken:
Great, I went ahead and merged this. Thanks!

LordOfXen:
My pleasure!

PR #657, opened 2025-11-14
cmake: add exports file for Apple

### LINKED ISSUE #656: provide an exported symbols list to linker for Apple Mach-O targets
As the $subject says, we should provide an exported symbols list to
linker for Apple Mach-O targets, in order to prevent prevent third-
party symbols from leaking into our library: such may be the case if
we statically link a vendored library to us, e.g. libjxl which leaks
libjxl and brotli symbols into SDL_image.
Reference issue: https://github.com/libsdl-org/libjxl/pull/2

We can generate an exports file at cmake configuration time from the
existing version script file, like below:

```cmake
set(sdlimage_exports)
file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/src/SDL_image.sym" sdlimage_exports_strings)
foreach(sdlimage_export_string ${sdlimage_exports_strings})
  string(STRIP "${sdlimage_export_string}" sdlimage_export_string)
  string(REGEX REPLACE "\;" "" sdlimage_export_string "${sdlimage_export_string}")
  if(sdlimage_export_string MATCHES "IMG_.*")
     # Prepend '_' to symbol name for Mach-O
     set(sdlimage_exports "${sdlimage_exports}_${sdlimage_export_string}\n")
  endif()
endforeach()
file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/SDL_image.exports" "${sdlimage_exports}")

# then add:
#
#    -Wl,-exported_symbols_list,'${CMAKE_CURRENT_BINARY_DIR}/SDL_image.exports'
#
# to linker flags
```

If we do this, we should do it similary in SDL_mixer and SDL_ttf, too.

Leaving implementation to @madebr

/cc @slouken
/cc @elevenfive

madebr:
- After adding a SDL_image symbol, you should run `src/genexports.py` to update `src/SDL_image.sym` and `src/SDL_image.exports`.
- The `genexports.py` script is satellite-library independent: it can be re-used for the other satellite libraries.
  It uses `.wikiheaders-options` to know the name of the project, the include directories and what regex to use for symbols.
- Consistency of all exports files is tested on ci

@slouken This needs to be hooked up in the Xcode project.

Link options should be:
```
-Wl,-exported_symbols_list,__PROJECT_DIR__/src/SDL_image.exports
```

Fixes #656

### DIFF
--- a/Xcode/SDL_image.xcodeproj/project.pbxproj
+++ b/Xcode/SDL_image.xcodeproj/project.pbxproj
@@ -734,6 +734,7 @@
 		007288120F0DA5BA00C302A9 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				EXPORTED_SYMBOLS_FILE = "$(SRCROOT)/../src/SDL_image.exports";
 				OTHER_LDFLAGS = "$(CONFIG_FRAMEWORK_LDFLAGS)";
 			};
 			name = Debug;
@@ -800,6 +801,7 @@
 		007288160F0DA5C400C302A9 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				EXPORTED_SYMBOLS_FILE = "$(SRCROOT)/../src/SDL_image.exports";
 				OTHER_LDFLAGS = "$(CONFIG_FRAMEWORK_LDFLAGS)";
 			};
 			name = Release;


sezero (approved):
Nice that this also caught miissing IMG_isANI and IMG_LoadANIAnimation_IO in SDL_image.sym.

src/genexports.py
```diff
@@ -0,0 +1,574 @@
...
+#     in a product, an acknowledgment in the product documentation would be
+#     appreciated but is not required.
+#  2. Altered source versions must be plainly marked as such, and must not be
+#     misrepresented as being the original software.
+#  3. This notice may not be removed or altered from any source distribution.
+
+# WHAT IS THIS?
+#  When you add a public API to a SDL library, please run this script, make sure the
+#  output looks sane (git diff, it adds to existing files), and commit it.
+#  It keeps the export lists in sync with tha API.
```
madebr:
```suggestion
#  It keeps the export lists in sync with the API.
```

sezero:
> @slouken This needs to be hooked up in the Xcode project.

It should be a two-liner, something like below (manual edit, _not tested_)

```diff
diff --git a/Xcode/SDL_image.xcodeproj/project.pbxproj b/Xcode/SDL_image.xcodeproj/project.pbxproj
index 0a9544e..ff606ab 100644
--- a/Xcode/SDL_image.xcodeproj/project.pbxproj
+++ b/Xcode/SDL_image.xcodeproj/project.pbxproj
@@ -680,6 +680,7 @@
 				DYLIB_CURRENT_VERSION = 301.0.0;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_TESTABILITY = YES;
+				EXPORTED_SYMBOLS_FILE = "$(SRCROOT)/../src/SDL_image.exports";
 				"FRAMEWORK_SEARCH_PATHS[sdk=appletv*]" = "\"$(PROJECT_DIR)/iOS\"";
 				"FRAMEWORK_SEARCH_PATHS[sdk=iphone*]" = "\"$(PROJECT_DIR)/iOS\"";
 				"FRAMEWORK_SEARCH_PATHS[sdk=macosx*]" = "\"$(PROJECT_DIR)/macOS\"";
@@ -747,6 +748,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 301.0.0;
 				DYLIB_CURRENT_VERSION = 301.0.0;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				EXPORTED_SYMBOLS_FILE = "$(SRCROOT)/../src/SDL_image.exports";
 				"FRAMEWORK_SEARCH_PATHS[sdk=appletv*]" = "\"$(PROJECT_DIR)/iOS\"";
 				"FRAMEWORK_SEARCH_PATHS[sdk=iphone*]" = "\"$(PROJECT_DIR)/iOS\"";
 				"FRAMEWORK_SEARCH_PATHS[sdk=macosx*]" = "\"$(PROJECT_DIR)/macOS\"";
```

madebr:
> ```diff
> EXPORTED_SYMBOLS_FILE = "$(SRCROOT)/../src/SDL_image.exports";
> ```

Not 100% sure it is correct, [looking at this output](https://github.com/madebr/SDL_image/actions/runs/19407451908/job/55524222992#step:8:1691).
Committing the code like this is perhaps the best way to get it fixed :)

sezero:
> Not 100% sure it is correct, [looking at this output](https://github.com/madebr/SDL_image/actions/runs/19407451908/job/55524222992#step:8:1691).

I can't pick the exact error among all the noise. I guess it's not due to the SDL version included in Xcode project is too old???

> Committing the code like this is perhaps the best way to get it fixed :)

Indeed: @slouken

slouken:
Please don't commit the code broken. I can take a look at this tomorrow.

elevenfive:
@madebr are you going to do another PR against 3.2.x branch?

madebr:
> @madebr are you going to do another PR against 3.2.x branch?

I think it's fine to apply to 3.2.x, but leaving the decision to slouken.

slouken:
I added the exported symbols list to the Xcode project. This change all looks good to me, and it seems good to go to the other satellite libraries and 3.2.x branches.

PR #659, opened 2025-11-23
Build with -Wno-unused-parameter

madebr:
Also remove all additional `(void)`s that silenced this warning.


sezero:
I'd understand not wanting to worry about the warning in the future, but what is the point of removing those silencers that have been there and are harmless? (Same for other similar pull requests in other libraries.)

madebr:
> I'd understand not wanting to worry about the warning in the future, but what is the point of removing those silencers that have been there and are harmless? (Same for other similar pull requests in other libraries.)

These `(void)`s were added to silence the `-Wunused-parameter` warning.
After disabling this warning, they are noise now and adding (very small) cognitive load.

slouken (approved):
Thank you!

PR #661, opened 2025-12-24
Trailing comma is embedded in SDL_ELF_NOTE_DLOPEN

madebr:


madebr:
Changed the commit message: use semicolon instead of comma

PR #662, opened 2026-01-01
cmake: support building SDL3_image with both STB and libpng

madebr:
Current CMake does not support building a SDL3_image using both stb and libpng: both libraries support the PNG file format.
However, only libpng supports APNGs.

This PR adds a `SDLIMAGE_PNG_LIBPNG` CMake option so you can build SDL3_image with libpng support, and also WIC/STB/IMAGEIO.

The pr also fixes up the files to be included in the MSVC release artifact.
See https://github.com/madebr/SDL_image/actions/runs/20646677630 to see how the release artifacts look like.

The dmg build fails because there are invalid build-machine specific paths in the xcode project:
```
2026-01-01T22:28:22.2694070Z error: lstat(/Users/valve/projects/SDL_image-experimental/external/libwebp/src/webp/mux.h): No such file or directory (2) (in target 'webp' from project 'webp')
```


slouken (approved):

build-scripts/release-info.json
```diff
@@ -196,7 +200,8 @@
         ],
         "lib/@<@ARCH@>@/optional": [
           "bin/libavif-16.dll",
-          "bin/libpng16-16.dll",
+          "bin/libpng16.dll",
+          "bin/zlib1.dll",
```
sezero: Is this for the case of vendored builds? If not, it's wrong: see the libpng dlls under msvc project which don't depend on zlib dll and are named libpng16-16.dll .

slouken:
The Xcode project has been fixed, thanks!

build-scripts/release-info.json
madebr: Yes.

The current CMake defaults are to build everything in shared mode, or in static mode, depending on `SDLIMAGE_DEPS_SHARED`.
You can change the sharedness of every dependency, but not for zlib specific.

sezero: OK then

madebr: Looking now into using a static zlib library.

sezero: Name difference (autotools vs cmake) is an annoyance too.

madebr: I think this is now addressed.
https://github.com/madebr/SDL_image/actions/runs/20648729927

There's one difference left: a zlib license is added to the arm builds.
Perhaps one should be added to `VisualC/external/optional/{x86,x64}` as well?

CMakeLists.txt
```diff
@@ -128,6 +128,7 @@ option(SDLIMAGE_WEBP "Support loading WEBP images" ON)
 option(SDLIMAGE_XCF "Support loading XCF images" ON)
 option(SDLIMAGE_XPM "Support loading XPM images" ON)
 option(SDLIMAGE_XV "Support loading XV images" ON)
+cmake_dependent_option(SDLIMAGE_PNG_LIBPNG "Support loading PNGs using libpng (useful for APNGs)" ON SDLIMAGE_PNG OFF)
```
sezero, line 131: Is this truly necessary? IMO, it just complicates things..

build-scripts/release-info.json
sezero: > There's one difference left: a zlib license is added to the arm builds. Perhaps one should be added to `VisualC/external/optional/{x86,x64}` as well?

Necessary? Isn't libpng zlib-licensed? I'm certainly not a lawyer, though..

CMakeLists.txt
madebr: You want SDL3_image.dll (without access to external libraries) to support pngs/jpegs using stbimage/wic/imageio. So `SDLIMAGE_PNG=ON` should be on.
If you want to load a APNG with the same SDL3_image.dll, you can then add `libpng16-16.dll` to the PATH and support those.

I'm open to suggestions for simplifying this.

build-scripts/release-info.json
madebr: Not necessary maybe, but as a courtesy?

CMakeLists.txt
sezero: Apologies.. I actually meant that for SDLIMAGE_ZLIB_SHARED, but commented at the wrong place...

CMakeLists.txt
```diff
@@ -217,11 +218,7 @@ if(SDLIMAGE_VENDORED AND SDLIMAGE_PNG_VENDORED)
 else()
     set(SDLIMAGE_ZLIB_VENDORED OFF)
 endif()
-if(SDLIMAGE_PNG_SHARED)
-    set(SDLIMAGE_ZLIB_SHARED ON)
-else()
-    set(SDLIMAGE_ZLIB_SHARED OFF)
-endif()
+cmake_dependent_option(SDLIMAGE_ZLIB_SHARED "Build zlib as a shared library" "${SDLIMAGE_PNG_SHARED}" SDLIMAGE_ZLIB_VENDORED OFF)
```
sezero: Is this truly necessary? IMO, it just complicates things..

build-scripts/release-info.json
sezero: OK by me either way.

CMakeLists.txt
madebr: I was afraid you could end up with multiple copies of zlib on Linux and get conflicts.

sezero: > I was afraid you could end up with multiple copies of zlib on Linux and get conflicts.

I'm saying that just lose the option and always build it statically already.

madebr: That's fine with me. It's easier to add an option if needed then remove one.

sezero (approved):

PR #666, opened 2026-01-06
Cleaned up PNG loading logic

slouken:
If libpng is enabled and available, use that for PNG loading. Otherwise, use WIC and ImageIO if they've been enabled, otherwise use SDL's STB implementation.

Note that we always fall back to SDL's implementation so that if PNG support is enabled we'll always have an available decoder.

### DIFF
--- a/src/IMG_libpng.c
+++ b/src/IMG_libpng.c
@@ -732,7 +732,7 @@ typedef struct
 } DecompressionContext;
 
 static SDL_Surface *decompress_png_frame_data(DecompressionContext *context, png_bytep compressed_data, png_size_t compressed_size,
-                                              int width, int height, int png_color_type, int bit_depth, SDL_Color *palette_colors, int palette_count)
+                                              int width, int height, int png_color_type, int bit_depth, png_bytep chunk_PLTE, Uint32 size_PLTE, png_bytep chunk_tRNS, Uint32 size_tRNS)
 {
     /*
      * Usually you'd directly decompress zlib but then we have to do defiltering and deinterlacing ourselves.
@@ -745,13 +745,11 @@ static SDL_Surface *decompress_png_frame_data(DecompressionContext *context, png
     // Create a memory stream to hold our synthetic PNG
     context->mem_stream = SDL_IOFromDynamicMem();
     if (!context->mem_stream) {
-        SDL_SetError("Failed to create memory stream");
         goto error;
     }
 
     // Write PNG signature
     if (SDL_WriteIO(context->mem_stream, png_sig, 8) != 8) {
-        SDL_SetError("Failed to write PNG signature");
         goto error;
     }
 
@@ -762,7 +760,6 @@ static SDL_Surface *decompress_png_frame_data(DecompressionContext *context, png
 
         // Write IHDR length and type
         if (SDL_WriteIO(context->mem_stream, ihdr_header, 8) != 8) {
-            SDL_SetError("Failed to write IHDR header");
             goto error;
         }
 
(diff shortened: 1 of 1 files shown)


madebr:
The STB backend in SDL_image is only used for jpeg, yes? (stb is part of core SDL 3.4 now)
https://github.com/libsdl-org/SDL_image/blob/f6843cb389ed4f36c2af896860b0073f15fd6f79/CMakeLists.txt#L109

madebr (approved):
Super change! I can't say no to simplifications.

What do you think about this change to also report apng at the end of CMake configuration?
```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,7 +106,7 @@ cmake_dependent_option(SDLIMAGE_SAMPLES_INSTALL "Install the SDL3_image sample p
 option(SDLIMAGE_TESTS "Build unit tests?" OFF)
 option(SDLIMAGE_TESTS_INSTALL "Install unit tests?" OFF)
 
-option(SDLIMAGE_BACKEND_STB "Use stb_image for loading JPEG and PNG files" ON)
+option(SDLIMAGE_BACKEND_STB "Use stb_image for loading JPEG files" ON)
 cmake_dependent_option(SDLIMAGE_BACKEND_WIC "Add WIC backend (Windows Imaging Component)" OFF WIN32 OFF)
 cmake_dependent_option(SDLIMAGE_BACKEND_IMAGEIO "Use native Mac OS X frameworks for loading images" ON APPLE OFF)
 
@@ -822,8 +822,10 @@ if(SDLIMAGE_PCX)
 endif()
 
 list(APPEND SDLIMAGE_BACKENDS PNG)
+list(APPEND SDLIMAGE_BACKENDS APNG)
 set(SDLIMAGE_PNG_ENABLED FALSE)
 if(SDLIMAGE_PNG)
+    set(SDLIMAGE_APNG_ENABLED FALSE)
     set(SDLIMAGE_PNG_ENABLED TRUE)
     set(SDLIMAGE_LIBPNG_ENABLED FALSE)
 
@@ -876,6 +878,7 @@ if(SDLIMAGE_PNG)
     endif()
 
     if(SDLIMAGE_LIBPNG_ENABLED)
+      set(SDLIMAGE_APNG_ENABLED TRUE)
         target_compile_definitions(${sdl3_image_target_name} PRIVATE
             SDL_IMAGE_LIBPNG
         )
```

slouken:
> The STB backend in SDL_image is only used for jpeg, yes? (stb is part of core SDL 3.4 now)
> 
> https://github.com/libsdl-org/SDL_image/blob/f6843cb389ed4f36c2af896860b0073f15fd6f79/CMakeLists.txt#L109

Yep, that's correct.

slouken:
> What do you think about this change to also report apng at the end of CMake configuration?

I like it! Would you like to push it to this PR, or want me to?

madebr:
> > What do you think about this change to also report apng at the end of CMake configuration?
> 
> I like it! Would you like to push it to this PR, or want me to?

I just pushed it. No need to manually apply the patch.

CMakeLists.txt
```diff
@@ -876,6 +878,7 @@ if(SDLIMAGE_PNG)
     endif()
 
     if(SDLIMAGE_LIBPNG_ENABLED)
+      set(SDLIMAGE_APNG_ENABLED TRUE)
```
slouken: Spacing doesn't match here...

madebr: Oops. Thanks for checking things.

slouken: You're welcome!

PR #669, opened 2026-01-08
cmake: check -Wl,-undefined,error before use

### LINKED ISSUE #568: AppleClang warning when linking
This is super minor, but this happens on macOS 15.5 (possibly earlier), where the ld included in the Apple toolchain is unhappy about the -undefined error:
```
ld: warning: -undefined error is deprecated
```
E.g. can see it [here](https://github.com/libsdl-org/SDL_image/actions/runs/15798166841/job/44532938531#step:12:69) (happens with SDL3_mixer/SDL3_ttf also since they share the same file).

Here's the offending part:
https://github.com/libsdl-org/SDL_image/blob/925c19db4d0bc9809fd3ac25c7e0ef771b668390/cmake/PrivateSdlFunctions.cmake#L285-L287

I noticed that the CMakeLists.txt for SDL checks the linker flag before using it:
https://github.com/libsdl-org/SDL/blob/343ad3eddda6df6ba2cef1321299e75168e832a1/CMakeLists.txt#L629-L633

So probably just need something similar moved over @madebr

madebr:
Fixes #568


sezero:
```
-- Performing Test HAVE_WL_UNDEFINED_ERROR
-- Performing Test HAVE_WL_UNDEFINED_ERROR - Success
```

You should temporarily use an error-on-warning linker flag (I don't know what it is / if there is.)

Or, just remove usage of this linker flag.

madebr:
> You should temporarily use an error-on-warning linker flag (I don't know what it is / if there is.)
> 
> Or, just remove usage of this linker flag.

`-Wl,--fatal-warnings` seems to be gnu only. I removed the link option completely. If somebody knows the link-warning-as-error link option for Apple, please let me know.

sezero:
>  I removed the link option completely.

That'd be best.

Needs doing in all satellite libs and in SDL3 itself, as well.

sezero:
> Needs doing in all satellite libs and in SDL3 itself, as well.

... compat libraries probably need your recent changes too.

Issue #468, opened 2024-09-07
On Mac, rotating image in Finder does not result in rotated image after loading


coinbump:
Here's the code:

```SDL_Surface* surface = IMG_Load(path.string().c_str());```

This works usually, but if I use the Mac Finder to rotate an image (command-L), the image appears rotated in the Finder, but loads with the original orientation in SDLImage.

I suspect this is some kind of optimization in MacOS to make rotation a change in metadata instead of actually rotating the image pixels to save time, but I'm not sure.

slouken:
Can you attach a link to a rotated image?

coinbump:
![heart-full](https://github.com/user-attachments/assets/15efed3e-0536-4a11-9736-e980f0c1b03a)

Here is a simple pixel heart. It was originally oriented like a normal heart. Then I rotated it in the Finder. Photoshop opens it as rotated, but SDLImage does not. Hopefully uploading to GitHub here doesn't change it.

coinbump:
[heart-full.png.zip](https://github.com/user-attachments/files/16924718/heart-full.png.zip)

Just in case uploading alters the image somehow, here it is compressed.

madebr:
What happens if you remove the `__MACOSX` folder next to the image?

coinbump:
I don't see a __MACOSX folder, that might just be added when you zip it.

Interestingly the image appears rotated in both Finder and Photoshop, but non-rotated in the GitHub app.

So this may be some edge case that Adobe optimized for since they're Mac friendly.

This issue isn't the end of the world, just confusing since I assumed I could rotate images in the Finder and use with SDLImage, but I guess not. Easy enough to rotate elsewhere.

Might just want to make a note of it in the documentation somewhere.

<img width="930" alt="Screenshot 2024-09-09 at 6 23 55 AM" src="https://github.com/user-attachments/assets/48b111dc-dfbc-445a-8ad3-6f5063425a90">

coinbump:
<img width="723" alt="Screenshot 2024-09-09 at 6 31 50 AM" src="https://github.com/user-attachments/assets/672c819e-d9ca-4589-9372-2535edba37cf">

Ok, I figured it out. I opened the png in an app that shows the png metadata.

And the rotated image has an Orientation property that non-rotated pngs do not.

So I guess it would be up to SDLImage to rotate the png? Or maybe at least send us the metadata for orientation so we can do it ourselves.

madebr:
[...]

PR #672, opened 2026-01-08
Handle orientation metadata

slouken:
This supports the orientation in TIFF files and the XMP orientation metadata in PNG files when using libpng. It does not currently support Exif metadata.

Fixes https://github.com/libsdl-org/SDL_image/issues/468


sezero:
Pushed two follow-up fixes to this for libpng

slouken:
> Pushed two follow-up fixes to this for libpng

These broke the build:
https://github.com/libsdl-org/SDL_image/actions/runs/20837362220/job/59864684968?pr=673

sezero:
> > Pushed two follow-up fixes to this for libpng
> 
> These broke the build: https://github.com/libsdl-org/SDL_image/actions/runs/20837362220/job/59864684968?pr=673

Fixed

slouken:
> Fixed

Thanks!

PR #675, opened 2026-01-11
Add helpers for loading GPU textures (closes #546)

### LINKED ISSUE #546: Feature Request: Load image directly into SDL GPU texture
Hey there,

Since SDL_ttf has support the GPU API, does it make sense to add some convenience functions here for loading an image into a GPU texture?

Thanks

jsoulier:
There's no docs yet. I'll do another pass if we're okay with this

### DIFF
--- a/examples/showgpuimage.c
+++ b/examples/showgpuimage.c
@@ -26,8 +26,8 @@
 static SDL_Window *window;
 static SDL_GPUDevice *device;
 static SDL_GPUTexture *texture;
-static Uint32 texture_width;
-static Uint32 texture_height;
+static int texture_width;
+static int texture_height;
 
 static const char *get_file_path(const char *file)
 {
--- a/src/IMG_gpu.c
+++ b/src/IMG_gpu.c
@@ -21,7 +21,7 @@
 
 #include <SDL3_image/SDL_image.h>
 
-static SDL_GPUTexture * LoadGPUTexture(SDL_GPUDevice *device, SDL_GPUCopyPass *copy_pass, SDL_Surface *surface, Uint32 *width, Uint32 *height)
+static SDL_GPUTexture * LoadGPUTexture(SDL_GPUDevice *device, SDL_GPUCopyPass *copy_pass, SDL_Surface *surface, int *width, int *height)
 {
     if (width) {
         *width = 0;
@@ -77,8 +77,7 @@ static SDL_GPUTexture * LoadGPUTexture(SDL_GPUDevice *device, SDL_GPUCopyPass *c
     const int row_bytes = surface->w * 4;
     if (row_bytes == surface->pitch) {
         SDL_memcpy(dst, src, row_bytes * surface->h);
-    }
-    else {
+    } else {
         for (int y = 0; y < surface->h; y++) {
             SDL_memcpy(dst + y * row_bytes, src + y * surface->pitch, row_bytes);
         }


include/SDL3_image/SDL_image.h
```diff
@@ -363,6 +363,10 @@ extern SDL_DECLSPEC SDL_Texture * SDLCALL IMG_LoadTexture_IO(SDL_Renderer *rende
  */
 extern SDL_DECLSPEC SDL_Texture * SDLCALL IMG_LoadTextureTyped_IO(SDL_Renderer *renderer, SDL_IOStream *src, bool closeio, const char *type);
 
+extern SDL_DECLSPEC SDL_GPUTexture * SDLCALL IMG_LoadGPUTexture(SDL_GPUDevice *device, SDL_GPUCopyPass *copy_pass, const char *file, Uint32 *width, Uint32 *height);
+extern SDL_DECLSPEC SDL_GPUTexture * SDLCALL IMG_LoadGPUTexture_IO(SDL_GPUDevice *device, SDL_GPUCopyPass *copy_pass, SDL_IOStream* src, bool closeio, Uint32 *width, Uint32 *height);
+extern SDL_DECLSPEC SDL_GPUTexture * SDLCALL IMG_LoadGPUTextureTyped_IO(SDL_GPUDevice *device, SDL_GPUCopyPass *copy_pass, SDL_IOStream* src, bool closeio, const char* type, Uint32 *width, Uint32 *height);
```
madebr: These new symbols need to be added to `src/SDL_image.sym` and `src/SDL_image.exports`.

CMakeLists.txt
```diff
@@ -1155,6 +1156,11 @@ if(SDLIMAGE_XV)
     target_compile_definitions(${sdl3_image_target_name} PRIVATE LOAD_XV)
 endif()
 
+if(SDL_GPU)
```
madebr: SDL_Image cannot depend on CMake variables from the SDL3 project.
I think we can unconditionally enable GPU in SDL_image.

src/SDL_image.exports
```diff
@@ -10,6 +10,9 @@ _IMG_LoadBMP_IO
 _IMG_LoadCUR_IO
 _IMG_LoadGIFAnimation_IO
 _IMG_LoadGIF_IO
+_IMG_LoadGPUTexture
```
slouken: These should go at the end of the list.

src/IMG_gpu.c
```diff
@@ -0,0 +1,131 @@
...
+        *height = surface->h;
+    }
+
+    SDL_DestroySurface(surface);
+    SDL_ReleaseGPUTransferBuffer(device, transfer_buffer);
+
+    return texture;
+}
+
+#define CHECK_PARAMS \
```
slouken: Please expand these out below so it's obvious what parameters are being validated and they can be easily updated if the implementation changes.

src/IMG_gpu.c
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    void *data = SDL_MapGPUTransferBuffer(device, transfer_buffer, 0);
+    if (!data) {
+        SDL_DestroySurface(surface);
+        SDL_ReleaseGPUTexture(device, texture);
+        SDL_ReleaseGPUTransferBuffer(device, transfer_buffer);
+        return NULL;
+    }
+    SDL_memcpy(data, surface->pixels, surface->w * surface->h * 4);
```
slouken: This needs to handle the case where surface->pitch != w * 4

slouken:
In general this looks good. Feel free to address the review feedback and add function documentation and I'll go ahead and merge it.

Thanks!

src/SDL_image.exports
sezero: The sym file isn't updated,either. Please revert changes to this file, and run genexports.py

jsoulier:
Thanks for the feedback, should be good for another look

include/SDL3_image/SDL_image.h
```diff
@@ -363,6 +363,129 @@ extern SDL_DECLSPEC SDL_Texture * SDLCALL IMG_LoadTexture_IO(SDL_Renderer *rende
...
+ * \param width a pointer filled in with the width of the GPU texture. may be NULL.
+ * \param height a pointer filled in with the width of the GPU texture. may be NULL.
+ * \returns a new GPU texture, or NULL on error.
+ *
+ * \since This function is available since SDL_image 3.4.0.
+ *
+ * \sa IMG_LoadGPUTextureTyped_IO
+ * \sa IMG_LoadGPUTexture_IO
+ */
+extern SDL_DECLSPEC SDL_GPUTexture * SDLCALL IMG_LoadGPUTexture(SDL_GPUDevice *device, SDL_GPUCopyPass *copy_pass, const char *file, Uint32 *width, Uint32 *height);
```
slouken: The type of surface->w and h are int, these types should match.

src/IMG_gpu.c
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    Uint8 *dst = SDL_MapGPUTransferBuffer(device, transfer_buffer, 0);
+    if (!dst) {
+        SDL_DestroySurface(surface);
+        SDL_ReleaseGPUTexture(device, texture);
+        SDL_ReleaseGPUTransferBuffer(device, transfer_buffer);
+        return NULL;
+    }
+    Uint8 *src = surface->pixels;
```
slouken:
```suggestion
    const Uint8 *src = surface->pixels;
```

src/IMG_gpu.c
```diff
@@ -0,0 +1,145 @@
...
+    const int row_bytes = surface->w * 4;
+    for (int y = 0; y < surface->h; y++) {
```
slouken: It's probably worth having the old code as a fast path when surface->pitch == surface->w * 4.

include/SDL3_image/SDL_image.h
jsoulier: I had it as Uint32 because that's what AcquireGPUSwapchainTexture uses. Fine either way

src/IMG_gpu.c
```diff
@@ -0,0 +1,145 @@
...
+    if (!device) {
+        SDL_InvalidParamError("device");
+        return NULL;
+    }
+    if (!copy_pass) {
+        SDL_InvalidParamError("copy_pass");
+        return NULL;
+    }
+
+    return LoadGPUTexture(device, copy_pass, IMG_Load_IO(src, closeio), width, height);
```
slouken:
```suggestion
    return IMG_LoadGPUTextureTyped_IO(device, copy_pass, src, closeio, NULL, width, height)
```
Maybe simpler as this?

examples/showgpuimage.c
```diff
@@ -0,0 +1,70 @@
...
+{
+    SDL_ReleaseGPUTexture(device, texture);
+    SDL_GPUCopyPass *copy_pass = SDL_BeginGPUCopyPass(command_buffer);
+    texture = IMG_LoadGPUTexture(device, copy_pass, path, &texture_width, &texture_height);
+    SDL_EndGPUCopyPass(copy_pass);
+}
+
+int main(int argc, char *argv[])
+{
+    SDL_Init(SDL_INIT_VIDEO);
```
slouken: Need to add error checking. It would also be nice to accept the image file on the command line. Please check out showimage.c for an example of one way to do this. If you have to pick one option, the command line is preferred over drag and drop.

include/SDL3_image/SDL_image.h
slouken: Yeah, I see that. The rest of the image API doesn't use Uint32 though, so let's keep it int for consistency.

src/IMG_gpu.c
```diff
@@ -0,0 +1,141 @@
...
+        SDL_ReleaseGPUTexture(device, texture);
+        SDL_ReleaseGPUTransferBuffer(device, transfer_buffer);
+        return NULL;
+    }
+    const Uint8 *src = surface->pixels;
+    const int row_bytes = surface->w * 4;
+    if (row_bytes == surface->pitch) {
+        SDL_memcpy(dst, src, row_bytes * surface->h);
+    }
+    else {
```
slouken:
```suggestion
    } else {
```

madebr:
The android, visual studio, and xcode projects need to be modified as well. (Preferably pre-release :) )

jsoulier:
> The android, visual studio, and xcode projects need to be modified as well. (Preferably pre-release :) )

Is that something I should do? I wouldn't know what to do for xcode/android besides copy pasting the project files and swapping the names. I don't have a mac on hand

madebr:
> > The android, visual studio, and xcode projects need to be modified as well. (Preferably pre-release :) )
> 
> Is that something I should do? I wouldn't know what to do for xcode/android besides copy pasting the project files and swapping the names. I don't have a mac on hand

I can't create those either for the same reasons. I posted this message as a reminder because I read a 3.4 prerelease is imminent.

slouken:
Yep, I can update the projects after I merge this.

PR #678, opened 2026-01-16
Updated license of ImageIO loading code to Zlib

slouken:
This change is made with the written permission of Eric Wing


ewmailing:
This is Eric Wing. I consent to these changes.

PR #679, opened 2026-01-17
Fix test failures when installed in a package and run from there

smcv:
* test: Install rgbrgb.ini as a resource file, too
    
    Otherwise the corresponding test will fail when run "as-installed", for
    example in Debian's autopkgtest framework.

* testanimation: Load webp data from test directory if necessary
    
    When running "as-installed" tests from /usr/libexec/installed-tests, the
    resource files are not necessarily in the current working directory.


slouken:
Merged, thanks!

PR #685, opened 2026-01-24
cmake: use PNG::PNG_shared if available for SDLIMAGE_PNG_SHARED

### LINKED ISSUE #683: CMAKE: PNG fails on MinGW-w64
I tried to build `SDL3_image-3.4.0` with MinGW-w64.
When configuring with CMake, this error message is printed on the console:
```
CMake Error at cmake/PrivateSdlFunctions.cmake:220 (message):
  PNG::PNG is not a shared library, but has type=INTERFACE_LIBRARY
  CMakeLists.txt:906 (target_get_dynamic_library)
```
All my cross compilers for x86_64/i686/aarch64 show the same result.
The bug doesn't happen with version 3.2.6.
At the moment, I added the option `-DSDLIMAGE_PNG_LIBPNG=OFF` to CMake for bypassing the problem
However, I have not understood if this is an expected behaviour and this is the right solution for closing the issue.
I also tried to use `libpng-1.6.47` and `libpng-1.6.54` but it does not seem to be related to the version of the library.

madebr:
Fixes #683

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -894,13 +894,18 @@ if(SDLIMAGE_PNG)
         )
         if(SDLIMAGE_PNG_SHARED)
             if(NOT DEFINED SDLIMAGE_DYNAMIC_PNG)
+                if(TARGET PNG::png_shared)
+                    set(png_target PNG::png_shared)
+                else()
+                    set(png_target PNG::PNG)
+                endif()
                 target_include_directories(${sdl3_image_target_name} PRIVATE
-                    $<TARGET_PROPERTY:PNG::PNG,INCLUDE_DIRECTORIES>
-                    $<TARGET_PROPERTY:PNG::PNG,INTERFACE_INCLUDE_DIRECTORIES>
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -894,13 +894,18 @@ if(SDLIMAGE_PNG)
         )
         if(SDLIMAGE_PNG_SHARED)
             if(NOT DEFINED SDLIMAGE_DYNAMIC_PNG)
+                if(TARGET PNG::png_shared)
+                    set(png_tgt PNG::PNG_shared)
```
slouken:
```suggestion
                    set(png_target PNG::PNG_shared)
```
We're not constrained on variable size and this makes it more clear what's happening.

sezero: This is probably a stupid question, but, is the intention here `PNG::PNG_shared` or `PNG::png_shared`? (Note the case difference of `png`.)

madebr: Not a dumb question at all. It must all be lower cased. Thanks!

madebr: True. I renamed that variable.

PR #691, opened 2026-02-10
IMG_libpng: adjust restrict attribute for png_noconst15_structrp

6ziv:
change the defination of png_noconst15_structrp to match the __restrict attribute.

When trying to build SDL_image with libpng18, I got warnings on L239 and some other lines complaining on type mismatch. It seems that png_const_structrp does not match png_noconst15_structrp, because the latter misses the restrict attribute. And in my building environment, changing png_const_structp to png_const_structrp fixes the warning.

### DIFF
--- a/src/IMG_libpng.c
+++ b/src/IMG_libpng.c
@@ -99,12 +99,12 @@ typedef png_structp png_noconst15_structrp;
 typedef png_inforp png_noconst15_inforp;
 typedef png_const_inforp png_noconst16_inforp;
 #else
-typedef png_const_structp png_noconst15_structrp;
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

sezero:
I couldn't get a warning when I configured SDL3_image with `-DSDLIMAGE_DEPS_SHARED=OFF` on the cmake command line. Tested with libpng versions 1.2, 1.4, 1.5, 1.6, and 1.8 (1.8 built from libpng git libpng18 branch.) What compiler is this with? Can you post the exact warnings please, as well as how you configured SDL3_image?

PR #697, opened 2026-02-18
cmake: fix `find_package(ZLIB)` incompatible with vendored `ZLIB`

madebr:
This fixes the following error:
```cmake
CMake Error at /usr/lib64/cmake/ZLIB/ZLIB-targets.cmake:42 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined: ZLIB::ZLIB

  Targets not yet defined: ZLIB::ZLIBSTATIC
```


sezero (approved):

sezero:
Is this one finalized?

nightmareci:
I've been waiting on a fix for this issue too. I've implemented my own fix locally in the mean time, using a different method: libtiff/libpng use `find_package(ZLIB)`, so I created a fake zlib package in the SDL_image repo and added it to the packages search path when vendored zlib is being used, then libtiff/libpng pick that up and proceed with using the SDL_image vendored zlib. My solution doesn't require any changes to the libtiff/libpng submodules.

madebr:
> libtiff/libpng use `find_package(ZLIB)`, so I created a fake zlib package in the SDL_image repo and added it to the packages search path when vendored zlib is being used, then libtiff/libpng pick that up and proceed with using the SDL_image vendored zlib. My solution doesn't require any changes to the libtiff/libpng submodules.

I like your solution more. Can you clean it up and create a pr?

nightmareci:
> > libtiff/libpng use `find_package(ZLIB)`, so I created a fake zlib package in the SDL_image repo and added it to the packages search path when vendored zlib is being used, then libtiff/libpng pick that up and proceed with using the SDL_image vendored zlib. My solution doesn't require any changes to the libtiff/libpng submodules.
> 
> I like your solution more. Can you clean it up and create a pr?

Sure.

sezero:
I suggest that this one be merged instead of #705

madebr:
> I suggest that this one be merged instead of #705

Let's do that :)

PR #725, opened 2026-04-10
gif: Default loop count to -1 when NETSCAPE extension is absent

jsm174:
We use `SDL_image` for loading GIFs in Visual Pinball and need to know the loop count.  Since the GIFs are created by table creators, many of them are missing the NETSCAPE header. When this happens, the loop count is defaulted to `0` which means loop indefinitely. We could set it to `1` so that it plays once, but I think it's better to let the user decide and return `-1` when the loop count can't be determined.

```
 IMG_AnimationDecoder* decoder = IMG_CreateAnimationDecoder_IO(io, false, nullptr);
 if (decoder)
 {
    SDL_PropertiesID props = IMG_GetAnimationDecoderProperties(decoder);
    m_loopCount = static_cast<int>(SDL_GetNumberProperty(props, IMG_PROP_METADATA_LOOP_COUNT_NUMBER, -1));
    IMG_CloseAnimationDecoder(decoder);
 }
```

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -547,7 +547,7 @@ static bool IMG_AnimationDecoderGetGIFHeader(IMG_AnimationDecoder *decoder, char
     }
 
     if (loopCount) {
-        *loopCount = 0;
+        *loopCount = -1;
     }
 
     IMG_AnimationDecoderContext *ctx = decoder->ctx;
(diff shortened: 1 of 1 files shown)


slouken:
This can potentially change how images are played by applications, so I would consider this a protocol change. It also makes sense to me that if a GIF doesn't specify how many times to loop, then it would loop indefinitely, which is the normal behavior of GIFs in the wild.

Since you are presumably in control of the assets in your game, I think it makes sense you can either author them with the metadata that makes sense for your game, or you can treat the value 0 as "unspecified" and do whatever makes sense for your content.

slouken:
Re-reading your comment I see that you aren't in control of the content, so if you really want something like this, it probably makes sense to add a new property that specifies whether the extended header was present while parsing, and check that when you see the value 0.

jsm174:
Created https://github.com/libsdl-org/SDL_image/pull/726 to discuss further and provide a different flow.

PR #726, opened 2026-04-10
gif: fix incorrect loop_count calculation

jsm174:
This PR supersedes https://github.com/libsdl-org/SDL_image/pull/725 with new logic.

The GIF89a spec has no concept of looping. When the NETSCAPE extension is present, the stored value is the number of repeats after the first play.

As of now, if the NETSCAPE extension is not present, `loop_count` will be `0` which means users trying to use this value would think they need to loop indefinitely.

The value for the `IMG_PROP_METADATA_LOOP_COUNT_NUMBER` property could be: 

  - No NETSCAPE extension: loop_count = 1 (play once)
  - NETSCAPE repeat=0: loop_count = 0 (infinite)
  - NETSCAPE repeat=N: loop_count = N+1 (total plays)

This also updates the encoder to skip writing the NETSCAPE extension if loop_count=1, ie:

  - loop_count == 1: skip NETSCAPE extension entirely (play once)
  - loop_count == 0: write NETSCAPE with repeat=0 (infinite)
  - loop_count > 1: write NETSCAPE with repeat=loop_count-1

see [here](https://github.com/libsdl-org/SDL_image/commit/38b93a90599da351c3e67cc5729aa847c12d9b99), [here](https://github.com/libsdl-org/SDL_image/pull/603), and [here](https://github.com/mirrorer/giflib/blob/fa37672085ce4b3d62c51627ab3c8cf2dda8009a/doc/whatsinagif/animation_and_transparency.html#L306-L307)

(If this is not acceptable, and we need a new `IMG_PROP_METADATA` property, I can modify this commit. At the end of the day, I just need an accurate way to know if I need to loop, and if so, how many times.)

### DIFF
--- a/src/IMG_gif.c
+++ b/src/IMG_gif.c
@@ -547,7 +547,7 @@ static bool IMG_AnimationDecoderGetGIFHeader(IMG_AnimationDecoder *decoder, char
     }
 
     if (loopCount) {
-        *loopCount = 0;
+        *loopCount = 1;
     }
 
     IMG_AnimationDecoderContext *ctx = decoder->ctx;
@@ -639,7 +639,8 @@ static bool IMG_AnimationDecoderGetGIFHeader(IMG_AnimationDecoder *decoder, char
                                     return SDL_SetError("Error reading Netscape sub-block data");
                                 }
                                 if (sub_block_data[0] == 0x01 && loopCount) {
-                                    *loopCount = LM_to_uint(sub_block_data[1], sub_block_data[2]);
+                                    int repeat = LM_to_uint(sub_block_data[1], sub_block_data[2]);
+                                    *loopCount = (repeat == 0) ? 0 : repeat + 1;
                                 }
                                 // Terminator
                                 if (!ReadOK(src, &sub_block_size, 1) || sub_block_size != 0x00) {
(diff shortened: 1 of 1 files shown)


jsm174:
```
➜  ~ gifsicle --no-loopcount base.gif > 1.gif
➜  ~ grep -aob "NETSCAPE2.0" 1.gif
➜  ~ gifsicle --loopcount=1 base.gif > 2.gif
➜  ~ grep -aob "NETSCAPE2.0" 2.gif
22:NETSCAPE2.0
➜  ~ gifsicle --loopcount=2 base.gif > 3.gif
➜  ~ grep -aob "NETSCAPE2.0" 3.gif
22:NETSCAPE2.0
➜  ~ gifsicle --loopcount=0 base.gif > 0.gif
➜  ~ grep -aob "NETSCAPE2.0" 0.gif
22:NETSCAPE2.0
```

I guess some of the confusion is that it's not really loop count but "repeats after the first play".

With what I provided, SDL_image would now abstract this concept away. 

So I'm definitely fine with changing to IMG_PROP_METADATA_NETSCAPE_EXTENSION, just let me know what the name should be and what type of value it should be.

slouken:
I reviewed the animation support and agreed with your conclusion. I went ahead and made a similar change and fixed the AVIF loop count as well.

Thanks for digging into this!

jsm174:
@slouken - awesome. Thank you very much!

slouken:
You're welcome!

PR #728, opened 2026-04-19
Tests: Copy resources to binary directory after build

Sackzement:
The image files are not copied to the build directory when building the tests.

This means when trying to run the tests (`testimage` and `testanimation`) you have to either manually copy the image files to the build directory or you have to _install_ the tests(which _does_ copy the images to the install directory).

This commit adds a copy operation to the build phase too.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -98,12 +98,25 @@ function(add_sdl_image_test NAME)
     endif()
 endfunction()
 
+function(copy_sdl_image_test_resources TARGET)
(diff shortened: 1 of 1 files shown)


slouken:
This seems reasonable to me. @madebr, is there a reason we didn't do this?

madebr (approved):

madebr:
> This seems reasonable to me. @madebr, is there a reason we didn't do this?

This was an oversight.
CTest sets the working directory to the source directory, so the resources are always available, just not for manual invocations.

slouken:
Merged, thanks!

PR #732, opened 2026-04-25
Update SECURITY.md with vulnerability details

GHYoungKyun:
Document security vulnerability related to heap buffer overflow in APNG chunk parsing.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,151 @@
+## Summary
+
+A malformed APNG file can trigger a heap buffer overflow in core APNG chunk parsing logic:
+
+**Affected Version**: `release-3.4.2`
+**File**: `src/IMG_libpng.c`
+**Function**: `read_png_chunk(...)` via `IMG_CreateAPNGAnimationDecoder(...)`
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### 3) PoC file
+
+The trigger file uses a PNG/APNG chunk with an oversized length field so that:
+
+- `data_length` is attacker-controlled
+- `chunk_size` wraps in 32-bit arithmetic
+- the decoder allocates only a tiny heap buffer
+- `read_png_chunk()` writes past the end immediately
```
madebr, line 107: Please attach a "trigger file" to this PR.

sezero:
I guess this demonstrates how _not_ to report a vulnerability :(

Please at least attach your trigger file here.

/cc @LordOfXen @slouken

slouken:
We do not plan on adding SECURITY.md to the repo. However, we welcome bug reports. Please include the details of the issue in the report along with a proof of concept file that triggers the issue. If you have a potential fix, please include that as a patch or separate linked PR.

PR #736, opened 2026-04-30
Enhance SECURITY.md with heap buffer overflow vulnerability details

GHYoungKyun:
https://github.com/libsdl-org/SDL_image/pull/732

Thanks for your feedback.
I've attached a trigger file (.png) that reproduces the issue.

Let me know if you need additional details.


slouken:
This is fixed, thanks for the report and POC!

PR #739, opened 2026-04-30
fix #738: add ANDROID to vendored_default

### LINKED ISSUE #738: logic for vendored_default does not handle ANDROID
the logic for vendored_default in CMakeLists.txt is:
```
if(MSVC)
    set(vendored_default ON)
else()
    set(vendored_default OFF)
endif()
```

However, the logic for vendored_default in SDL_mixer and SDL_TTF is:
```
if(ANDROID OR MSVC)
    set(vendored_default ON)
else()
    set(vendored_default OFF)
endif()
```

It seems like SDL_image not handling ANDROID is an oversight.

bostick:


slouken:
Merged, thanks!

PR #743, opened 2026-05-03
[SDL2] Bump to libavif 1.1.1

madebr:


sezero:
Do we really need the libwebp upgrade to 1.3.2? If it's for sharpyuv, it's only used in `avifImageRGBToYUV` which is irrelevant in SDL2_image having no SAVE_AVIF option.

madebr:
> Do we really need the libwebp upgrade to 1.3.2? If it's for sharpyuv, it's only used in `avifImageRGBToYUV` which is irrelevant in SDL2_image having no SAVE_AVIF option.

No we don't. I blindly used the same commits as the SDL3 pr.
libwebp needed a patch for allowing sharpyuv use in libavif.

macos ci is failing because of sharpyuv and webp of SDL2 does not provide a sharpyuv target, so I now disable it by default, which makes it somewhat easier.

PR #747, opened 2026-05-25
Add CYGWIN support for DLL library check

carlo-bramini:
I tried to build SDL_image for CYGWIN, but it fails with the error:
```
CMake Error at cmake/PrivateSdlFunctions.cmake:212 (message):
  "cygavif-16.dll" is not a .so shared library
```
The bug is inside this `cmake/PrivateSdlFunctions.cmake`.
We must handle `CYGWIN` like `WIN32` because it also uses DLL files.
Attached patch fixes this tiny issue.


sezero (approved):
Looks OK to me. And I guess this is needed everywhere. @madebr?

madebr (approved):

madebr:
> Looks OK to me. And I guess this is needed everywhere. @madebr?

Yup, OP will need to "infect" all other satellite libraries with this change.

sezero:
Applied it everywhere.

sezero:
Do we want this in satellite libs' release branches?

madebr:
> Do we want this in satellite libs' release branches?

The latest SDL3 release does not support cygwin.
cygwin is currently only supported in the main branches.

PR #764, opened 2026-09-11
cmake: initialize SDLIMAGE_APNG_ENABLED unconditionally

GEvangelinos:
- [x] I confirm that I am the author of this code and release it to the SDL project under the zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

## Description
When configuring with `-DSDLIMAGE_PNG=OFF -Wdev`, CMake triggers an `AUTHOR_WARNING`:
```
CMake Warning (dev) at CMakeLists.txt:1374 (message):
  SDLIMAGE_APNG_ENABLED not defined
```

`APNG` is unconditionally appended to `SDLIMAGE_BACKENDS`, but `set(SDLIMAGE_APNG_ENABLED FALSE)` was nested inside `if(SDLIMAGE_PNG)`. When PNG is disabled, the reporting loop encounters an undefined variable.

Moving `set(SDLIMAGE_APNG_ENABLED FALSE)` outside the conditional ensures the variable is always defined.


madebr (approved):

madebr:
Merged, thanks for the pull request!

Pushed to 3.4.x as a324e90af4c649e3b9f980b7f45868182b494dd0

